PR #1, opened 2018-12-07
Unif mem

cosunae:
currently under self review


cosunae:
@wdeconinck all the gpu test work for me already, perhaps you can try to do a test with the openacc pgi?

PR #2, opened 2018-12-13
update to latest API of GT

cosunae:
Compatible with 
https://github.com/GridTools/gridtools/tree/16.10.2018.p1


wdeconinck:
Merged in develop.

PR #12, opened 2020-03-04
PointCloud Fortran interface

benjaminmenetrier:
This PR does two things:
 - removing the "No Connectivity" mesh used previously in JEDI,
 - adding a Fortran interface for the PointCloud function space, in order to replace the NodeColumns function spaces initialized with the "No Connectivity" mesh in JEDI.

Minimal interfaces with the C++ class `PointCloud` are implemented, more could be added later.


benjaminmenetrier:
Thanks @wdeconinck, working with you and the others this week was very nice and interesting!

Issue #41, opened 2020-04-20
Creating a distribution for a 40000x20000 latlon grid is too expensive


pmarguinaud:
Creating such a distribution requires several Gb of memory per MPI task and takes a long time (minutes) to get created, which is not acceptable for the application we work on.

PR #42, opened 2020-04-20
Support memoryfree distributions

pmarguinaud:
Solves issue #41 .

Allows creating a distribution using a Config object. This Config object is either used directly by the DistributionImpl class, or passed to the Partitioner class.

A nbands option was added to the CheckerBoard partitioner; this option makes it possible to ask for a partition with nbands stripes along the N/S direction. 

A light option was added to the DistributionImpl ctor; setting this option to true creates a trivial N/S partition. The blocksize option makes it possible to force the size of each MPI partition to be a multiple of this value (useful for lat/lon grids). When this light option is enabled, the big partition array is not allocated; instead (as this partitioning is very simple), we directly can compute the rank each point belongs to.

To be efficient, the partition method of the Distribution class had to be implemented in the header file, so that it be inlined. The setup of the functionspace StructuredColumns (which is the one we intend to use) has been modified so that it uses this accessor method, instead of requesting access to the big partition array.

A test case has been added to demonstrate the use of the light and nbands options.

### DIFF
--- a/src/atlas/grid/detail/distribution/DistributionImpl.h
+++ b/src/atlas/grid/detail/distribution/DistributionImpl.h
@@ -29,8 +29,6 @@ class Grid;
 namespace grid {
 class Partitioner;
 }
-}
-
 }  // namespace atlas
 
 namespace atlas {
--- a/src/atlas/grid/detail/distribution/DistributionImpl.h
+++ b/src/atlas/grid/detail/distribution/DistributionImpl.h
@@ -15,14 +15,20 @@
 
 #include "atlas/util/Object.h"
 #include "atlas/util/vector.h"
-
+#include "atlas/util/Config.h"
 #include "atlas/library/config.h"
+
+
+namespace eckit {
+class Parametrisation;
+}
 
 namespace atlas {
 
 class Grid;
 namespace grid {
 class Partitioner;
+}
 }
 
 }  // namespace atlas
@@ -32,9 +38,11 @@ namespace grid {
 
 class DistributionImpl : public util::Object {
 public:
+    using Config = atlas::util::Config;
     using partition_t = atlas::vector<int>;
 
     DistributionImpl( const Grid& );
+    DistributionImpl( const Grid&, const eckit::Parametrisation & );
 
     DistributionImpl( const Grid&, const Partitioner& );
 
@@ -44,15 +52,21 @@ public:
 
     virtual ~DistributionImpl();
 
-    int partition( const gidx_t gidx ) const { return part_[gidx]; }
+    int partition( const gidx_t gidx ) const 
+    { 
+      if (part_.size () > 0)
+        return part_[gidx];
+      idx_t iblock = gidx / blocksize_;
+      return (iblock * nb_partitions_) / nb_blocks_;
+    }
 
-    const partition_t& partition() const { return part_; }
+    const partition_t& partition() const { checkPartition (); return part_; }
 
     idx_t nb_partitions() const { return nb_partitions_; }
 
     operator const partition_t&() const { return part_; }
 
-    const int* data() const { return part_.data(); }
+    const int* data() const { checkPartition (); return part_.data(); }
 
     const std::vector<idx_t>& nb_pts() const { return nb_pts_; }
 
@@ -63,8 +77,27 @@ public:
 
     void print( std::ostream& ) const;
 
+    size_t footprint () const
+    {
+      return nb_pts_.size () * sizeof (nb_pts_[0]) + part_.size () * sizeof (part_[0]);
+    }
+
+
 private:
-    idx_t nb_partitions_;
+
+    void setupWithPartitioner (const Grid &, const Partitioner &);
+
+    void checkPartition () const
+    {
+      if (part_.size () == 0)
+        throw_Exception ("partition array of distribution is empty");
+    }
+    // For trivial partitionning
+    size_t gridsize_ = 0;
+    idx_t nb_partitions_ = 0;
+    size_t blocksize_ = 0;
+    idx_t nb_blocks_ = 0;
+
     partition_t part_;
     std::vector<idx_t> nb_pts_;
     idx_t max_pts_;
--- a/src/atlas/functionspace/detail/StructuredColumns_setup.cc
+++ b/src/atlas/functionspace/detail/StructuredColumns_setup.cc
@@ -180,13 +180,13 @@ void StructuredColumns::setup( const grid::Distribution& distribution, const eck
                     auto& __j_end         = thread_reduce_j_end[thread_num];
                     auto& __owned         = thread_reduce_owned[thread_num];
                     idx_t c               = begin;
-                    const auto& partition = distribution.partition();
+
                     for ( idx_t j = thread_j_begin; j < thread_j_end; ++j ) {
                         auto& __i_begin = thread_reduce_i_begin[j][thread_num];
                         auto& __i_end   = thread_reduce_i_end[j][thread_num];
                         bool j_in_partition{false};
                         for ( idx_t i = thread_i_begin[j]; i < thread_i_end[j]; ++i, ++c ) {
-                            if ( partition[c] == mpi_rank ) {
+                            if ( distribution.partition (c) == mpi_rank ) {
                                 j_in_partition = true;
                                 __i_begin      = std::min<idx_t>( __i_begin, i );
                                 __i_end        = std::max<idx_t>( __i_end, i + 1 );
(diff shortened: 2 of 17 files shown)


wdeconinck:
Hi @pmarguinaud Thank you for the PR.
I will take this and try to modify it a little bit to better match the Atlas philosophy, while still remaining the same footprint, performance of accessing `partition(idx)` and scheme.

pmarguinaud:
Thank you Willem. I was actually a bit worried about how you would react to this PR. 

It is, nevertheless, essential to my project.

pmarguinaud:
Hi Willem,

There is something wrong with the "light" distribution (for some number of MPI tasks).  Would you want me to look into that, or have decided to restart from scratch anyway ?

Regards

wdeconinck:
Please if you could look at that in current proposal that would definitely be helpful as the bug (?) will likely be similar in a further development otherwise. I have not yet had the chance to start on this yet.

pmarguinaud:
Hi Willem,

I have fixed the nb_pts calculation.

But I have found another problem (not related to the changes I introduced). I will look at it.

Philippe

wdeconinck:
Don't worry about the 1 failing travis test. Sometimes rerunning it fixes it. I'll try.
It is an intermittent MPI problem when I'm oversubscribing with 16 MPI tasks, only on OSX with OpenMPI (MPICH seems OK).
I quarantined it in latest develop branch yesterday, which is not yet in this branch.

pmarguinaud:
I still see something is not working. I am looking at it, but I would you to see what it is.

How do I submit code showing something is broken ?

wdeconinck:
If it is unrelated to this PR, you can create a test that fails in a draft PR perhaps?

src/atlas/grid/detail/distribution/DistributionImpl.h
```diff
@@ -44,15 +50,21 @@ class DistributionImpl : public util::Object {
...
     virtual ~DistributionImpl();
 
-    int partition( const gidx_t gidx ) const { return part_[gidx]; }
+    int partition( const gidx_t gidx ) const 
+    { 
+      if (part_.size () > 0)
+        return part_[gidx];
+      idx_t iblock = gidx / blocksize_;
+      return (iblock * nb_partitions_) / nb_blocks_;
+    }
```
wdeconinck, line 59: How important is inlining here?  I want to know if it could it be made virtual perhaps, preventing this to be inlined.

pmarguinaud: Hello Willem,

I have just tested it on my reference test case (40000x20000), and it costs only a few seconds, so yes, you should make it virtual.

Philippe

wdeconinck: Sometimes seconds matter... But typically this is only at a setup though, but could matter even there for pre- and post- processing programs.
If you believe it is OK, then that opens up a lot of options, and the "if" could also be removed.

pmarguinaud: Of course, I understand, please go on.

wdeconinck:
Thanks @pmarguinaud , In my personal github fork I have published my modifications after rebasing your changes to latest develop: https://github.com/ecmwf/atlas/compare/develop...wdeconinck:feature/distribution-function  I am not sure how easy it is for you to test this, but I want to make sure that the solution is adequate for your purpose.
The main difference is that now `DistrubutionImpl` is polymorphic, with 2 baseclasses:
- DistributionArray  (which is the behaviour of before)
- DistributionFunction (which is a base class for functional memory-free implementations)
From DistributionFunction we can then derive the class BandsDistribution, which implements your particular proposal for any blocksize.
The Distribution now no longer exposes the API to access the internal `vector` (now belonging to DistributionArray). Instead now,  we have *only* retain the `DistributionImpl` function
```cpp
virtual int partition(gidx_t) = 0;
```
which has now become virtual, and may/will lead to a performance penalty. It is however yet to be discovered how much, and I would be interested to see a confirmation besides your previous test, on how this affects your use-case with a full implementation.

Parallel to the  `BandsDistribution`, we now also have two specialized `Partitioners`, which when asked to return a distribution, would create a `Distribution` implemented by the `BandsDistribution`.
The two are: `EqualBandsPartitioner`, `RegularBandsPartitioner`. The first creates the `Bandsdistribution` with blocksize=1, whereas the last one will do introspection of the grid to be partitioned, and set the blocksize to either grid.nx() if the grid is a RegularGrid, or to 1 otherwise.

The proposed API for your use-case would now be e.g.
```cpp
auto grid        = Grid("L40000x20000");
auto partitioner = Partitioner("regular_bands");
auto dist        = Distribution(grid,partitioner);
...
// Usage
for( gidx_t n = 0, size = dist.size(); n < size; ++n ) {
    int p = dist.partition( n );  // virtual call unfortunately
}
```

pmarguinaud:
Hello Willem,

I have cloned your branch and tried it on 4 AMD nodes with my 40000x20000 grid, and the cost is about 8 seconds (compared to 2 seconds) without direct inlining. 

So it looks OK to me. 

Philippe

wdeconinck:
> 
> 
> Hello Willem,
> 
> I have cloned your branch and tried it on 4 AMD nodes with my 40000x20000 grid, and the cost is about 8 seconds (compared to 2 seconds) without direct inlining.
> 
> So it looks OK to me.
> 
> Philippe

Out of interest, the cost of what exactly?
I also timed about 9 seconds on my workstation to iterate over all partitions with 4 MPI tasks (does not matter, as they all performed the loop), **without** OpenMP. I imagine that on a AMD node you could have a lot of threads available to reduce the cost to a fraction of this.

pmarguinaud:
I just created a 40000x20000 grid/distribution and its associated StructuredColumns functionspace.

wdeconinck:
Hi Philippe,
I have evolved the branch a little bit more, and received a factor 2x for the case "L40000x20000" and "regular_bands". 
Because there the blocksize is so large, there will not be any integer overflow happening, and we can use 32bit integers.

I have done a comparison of inline versus direct and did not see a big difference. I have now also added an API that can copy (using inline access internally) blocks of computed values to an array (e.g. a full latitude):
```cpp
gidx_t n = 0;
std::vector<int> partition( grid.nxmax() );
for( idx_t j = 0; j < grid.ny(); ++j, n += grid.nx( j ) ) {
    dist.partition( n, n + grid.nx( j ), part );
    for( idx_t i = 0; i < grid.nx( j ); ++i ) {
        int p = part[i];
    }
}
```
It seems to improve a little bit upon full virtual access. My new timings that I received for looping over all grid points and computing the partition with 32bit integers are:
```
inline:     3.522s
virtual:    4.118s
blocked:    4.021s

// On desktop workstation, single-threaded, gnu6.3, optimisation -O2 -g -DNDEBUG
```
I have not adapted the StructuredColumns yet to adopt the "blocked" approach. It may be a lot of work with added bugs for little benefit.

If it is not an issue for you to try your setup again with the new implementation, that would be helpful for reference. I am still suprised to see you go from 2s to 8s as these timings don't seem to indicate such large difference between inline and virtual... (but it is what it is I guess).

pmarguinaud:
Hello Willem,

I am sorry I could not answer earlier.

I ran the test again and noticed that last time I forgot to enable OpenMP. Here are the results on 4 AMD nodes, using 8 tasks/node and 16 threads/task :

```
10 : └─test regular_bands performance test │ 1   │  0.78157s │  0.78157s │  0.00000s │  0.78157s │  0.78157s │ test_distribution_regular_bands.cc +181
11 :   ├─inline access                     │ 1   │  0.02431s │  0.02431s │  0.00000s │  0.02431s │  0.02431s │ test_distribution_regular_bands.cc +207
12 :   ├─virtual access                    │ 1   │  0.12576s │  0.12576s │  0.00000s │  0.12576s │  0.12576s │ test_distribution_regular_bands.cc +214
13 :   └─virtual cached access             │ 1   │  0.13072s │  0.13072s │  0.00000s │  0.13072s │  0.13072s │ test_distribution_regular_bands.cc +221
```

I am very happy this works so well, with such a nice interface. I think this will take care of a weak point in Atlas. We could consider using function distribution for other kind of distribution, eg global 2.5km array distribution implies allocating 300Mb on each task, which is also quite expensive.

Cheers,

Philippe

wdeconinck:
OK I am glad that everything is perfect for you now.
Indeed for a long time I was quite worried about this global array. At least now there's something that can take us further, likely with some pro's and cons, in which distributions are achievable like that.

wdeconinck:
PR is now rebased and the discussed approach is now available in develop as of af652b1

Issue #43, opened 2020-04-20
The class atlas::vector lacks a copy constructor and an assignment operator


pmarguinaud:

wdeconinck:
I merged your PR, thank you!
I guess since PRs are treated like GH issues, we can simplify collaboration so that if you have a PR it can be used instead of a Issue.

PR #44, opened 2020-04-20
Copy constructor & assignment operator for atlas::vector

pmarguinaud:
Solves issue #43 .

A test has been added. 

a = b

The assignment operator creates a copy of the object to be copied (b), and its contents are swapped with the object being assigned (a). The temporary copy (which holds the contents of the object being assigned, ie a) is then deleted, which implies that the memory use by a is freed.

### DIFF
--- a/src/atlas/util/vector.h
+++ b/src/atlas/util/vector.h
@@ -37,10 +37,24 @@ public:
         assign( size, value );
     }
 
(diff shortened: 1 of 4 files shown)


wdeconinck:
Thank you @pmarguinaud for this PR!
I edited your commit and merged it in develop with commit  
37faae7to resolve conflicts and compilation. No API has been changed.

Issue #45, opened 2020-04-23
Fortran interface for distribution object lacks a nbpts method


pmarguinaud:
This method returns the number of grid points held by each MPI task.

PR #46, opened 2020-04-23
Grid distribution nbpts

pmarguinaud:
Solves issue #45 .


wdeconinck:
Hi @pmarguinaud  Thank you for your PR
I would suggest that instead of involving fckit_mpi to allocate nb_pts, you could instead add another Fortran binding to "nb_partitions" and use that instead.

pmarguinaud:
I followed your recommendation and added the nb_partitions binding; this binding is now used to size the nb_pts array returned by the binding of the same name.

wdeconinck:
Merged with commit d2bdd84
Thank you!

PR #47, opened 2020-04-27
Fortran interfaces for Point, Geometry and KDTree

benjaminmenetrier:
Hello @wdeconinck.

As mentioned in https://github.com/ecmwf/fckit/pull/11, I have implemented Fortran interfaces for several classes in ATLAS to get rid of the geometry aspects in the JCSDA fork of FCKIT:
 - PointXY
 - PointXYZ
 - PointLonLat
 - Earth
 - UnitSphere

I have replaced the calls to the old FCKIT geometry functions with the new ATLAS ones in a branch of my code: all the tests are still passing!

Let me know when I can do the same with the KD-Tree.


wdeconinck:
@benjaminmenetrier  Do you have any use-case for the Point classes in Fortran?

benjaminmenetrier:
Hi @wdeconinck. Not really, but I thought it would make the calls to `Earth` and  `UnitSphere` methods cleaner. For instance, if I want to compute the distance between two lon/lat points on the `UnitSphere`, I do:
```
! Local variables
type(atlas_unitsphere) :: aunitsphere

! Create ATLAS UnitSphere
aunitsphere = atlas_unitsphere()

! Compute distance
dist = aunitsphere%distance(atlas_pointlonlat(lon_i,lat_i),atlas_pointlonlat(lon_f,lat_f))
```
The other solution would be to have the lon/lat of both points as arguments of `atlas_unitsphere%distance`, without using the `atlas_pointlonlat`:
```
! Compute distance
dist = aunitsphere%distance(lon_i,lat_i,lon_f,lat_f)
```
If you prefer this solution, I can remove the Point classes Fortran interfaces and change the `Earth` and `UnitSphere` interfaces arguments, no problem. Just let me know.

wdeconinck:
Please check what I just pushed in develop.

-  In particular, there is now the templated `atlas::util::KDTree<Payload,Point>` class.
    
    However likely what you want to interface is the `atlas::util::IndexKDTree`
    class which fixes the     template arguments of `atlas::util::KDTree`

- Rather than the `UnitSphere` I also made a `Geometry` abstraction with a (simple) factory:

    ```
    auto earth = Geometry("Earth");
    auto unit_sphere = Geometry("UnitSphere");
    auto sphere = Geometry( radius );
    ```
    The latter may be useful if you want to use a model which uses a different
    radius than the one hardcoded inside Atlas.

    I think functions that are not part of the Geometry abstraction should perhaps be left out of 
    the Fortran Interface.

- I am OK with the Point classes :)

benjaminmenetrier:
Hi @wdeconinck. I'm done with the Fortran interfaces for the Point, Geometry and IndexKDTree classes. I have implemented tests and everything looks good now.

I have used these new interfaces instead of the FCKIT ones in my code, and I get the exact same results (which is a strong confirmation, since my code is very sensitive to geometry and KDTree results).

Last minor issue: the `distance` returned by the KDTree in the `Value` object is the Euclidian distance, not the great-circle distance (given by `geometry.distance()` for instance). This can be misleading. In the current implementation, I perform the conversion from Euclidian to spherical distance in my own code, not in the interfaces. Do you think we should do it in the interfaces?

At JCSDA, we would like to tag a version of our containers at the end of the week. It would be nice if we could use ECMWF versions of eckit, fckit and atlas for this tagged version. So could you review this PR before Thursday? No worries if you can't though, we'll tag the current JCSDA branches.

benjaminmenetrier:
Hi @wdeconinck. The `atlas_fctest_kdtree` test fails with PGI.
I installed PGI on my workstation: I can reproduce the segmentation fault but I can't figure out where it's coming from. Can you help me?

benjaminmenetrier:
@wdeconinck, sorry for the mess with the last 3 commits. Everything is back on track. The `atlas_fctest_kdtree` test is still failing with PGI.

wdeconinck:
Thanks @benjaminmenetrier  I will have a look at the PGI.

w.r.t. the distance, it is computed by the member function "distance" of the `Point` class... which is also used to find the closest points.
**If** it would it be equivalent to use the great-circle-distance in the kd-tree search, one could argue that we can return such distance instead, but still keep the fast `Point::distance` for lookup

benjaminmenetrier:
Regarding the distance, I understand it is faster to use the Euclidian distance for internal KDTree calculations, and I don't mind returning it at the end. Maybe we should just add a comment somewhere to let future users know.

src/atlas_f/util/atlas_Geometry_module.F90
```diff
@@ -0,0 +1,179 @@
...
+subroutine atlas_Geometry__delete(this)
+  use atlas_Geometry_c_binding
+  class(atlas_Geometry), intent(inout) :: this
+  if ( .not. this%is_null() ) then
+    call atlas__Geometry__delete(this%CPTR_PGIBUG_A)
+  end if
+  call this%reset_c_ptr()
+end subroutine atlas_Geometry__delete
+
+subroutine Geometry__xyz2lonlat(this, x, y, z, lon, lat)
```
wdeconinck, line 100: Can you create another overload `geometry%xyz2lonlat( xyz, lonlat )`
where xyz is a size(3) array and lonlat a size(2) array ?

src/atlas_f/util/atlas_Geometry_module.F90
```diff
@@ -0,0 +1,179 @@
...
+  class(atlas_Geometry), intent(in) :: this
+  real(c_double), intent(in) :: x
+  real(c_double), intent(in) :: y
+  real(c_double), intent(in) :: z
+  real(c_double), intent(out) :: lon
+  real(c_double), intent(out) :: lat
+  call atlas__Geometry__xyz2lonlat(this%CPTR_PGIBUG_A, x, y, z, lon, lat)
+end subroutine Geometry__xyz2lonlat
+
+subroutine Geometry__lonlat2xyz(this, lon, lat, x, y, z)
```
wdeconinck, line 112: Can you create another overload geometry%lonlat2xyz( lonlat, xyz )
where xyz is a size(3) array and lonlat a size(2) array ?

src/atlas_f/util/atlas_Geometry_module.F90
```diff
@@ -0,0 +1,179 @@
...
+  class(atlas_Geometry), intent(in) :: this
+  real(c_double), intent(in) :: lon
+  real(c_double), intent(in) :: lat
+  real(c_double), intent(out) :: x
+  real(c_double), intent(out) :: y
+  real(c_double), intent(out) :: z
+  call atlas__Geometry__lonlat2xyz(this%CPTR_PGIBUG_A, lon, lat, x, y, z)
+end subroutine Geometry__lonlat2xyz
+
+function Geometry__distance_lonlat(this, lon1, lat1, lon2, lat2) result(distance_lonlat)
```
wdeconinck, line 124: Can you create another overload `distance_lonlat( this, p1, p2 )`
where p1 and p2 are size(2) arrays?

src/atlas_f/util/atlas_Geometry_module.F90
```diff
@@ -0,0 +1,179 @@
...
+  class(atlas_Geometry), intent(in) :: this
+  real(c_double), intent(in) :: lon1
+  real(c_double), intent(in) :: lat1
+  real(c_double), intent(in) :: lon2
+  real(c_double), intent(in) :: lat2
+  real(c_double) :: distance_lonlat
+  distance_lonlat = atlas__Geometry__distance_lonlat(this%CPTR_PGIBUG_A, lon1, lat1, lon2, lat2)
+end function Geometry__distance_lonlat
+
+function Geometry__distance_xyz(this, x1, y1, z1, x2, y2, z2) result(distance_xyz)
```
wdeconinck, line 136: Can you create another overload `distance_xyz( p1, p2 )
where p1 and p2 are size(3) arrays?

src/tests/util/fctest_geometry.F90
```diff
@@ -0,0 +1,170 @@
...
+  lonlat1 = (/-71.6_c_double, -33._c_double/)
+  lonlat2 = (/121.8_c_double, 31.4_c_double/)
+  xyz1 = (/2.647e-1_c_double, -7.958e-1_c_double, -5.446e-1_c_double/)
+  xyz2 = (/-4.498e-1_c_double, 7.254e-1_c_double, 5.210e-1_c_double/)
+
+  ! Check conversion from cartesian to spherical
+  call geometry%xyz2lonlat(xyz1(1), xyz1(2), xyz1(3), lonlat(1), lonlat(2))
+  FCTEST_CHECK_CLOSE( lonlat(1) , lonlat1(1), 1.e-2_c_double )
+  FCTEST_CHECK_CLOSE( lonlat(2) , lonlat1(2), 1.e-2_c_double )
+
```
wdeconinck, line 68: Please add tests here and below with additional requested functions in atlas_Geometry_module

src/tests/util/fctest_kdtree.F90
```diff
@@ -0,0 +1,233 @@
...
+  deallocate(distances)
+  deallocate(lons_rad)
+  deallocate(lats_rad)
+  deallocate(indices_rad)
+  deallocate(distances_rad)
+  deallocate(result_lons)
+  deallocate(result_lats)
+  deallocate(result_indices)
+  deallocate(result_distances)
+
```
wdeconinck, line 228: Please add more tests with requested additional functions in IndexKDTree

benjaminmenetrier:
Hi @wdeconinck, I have added the interfaces and tests that you suggested. I have even added new interfaces with `lonlat` as a `size(2)` array for the  `insert` and `build` methods of the KD-Tree.

wdeconinck:
Hi @benjaminmenetrier I have already added the KDTree and Geometry to the develop branch

It seems to be independent of the Point.
Do you still have a use for it? The implementation is not very efficient but may still be useful.

benjaminmenetrier:
Hi @wdeconinck, thank you for merging the KDTree and Geometry into the `develop` branch. I don't use the `Point` class interface, so you can merge it if you think it might be useful eventually or discard it if you prefer.

wdeconinck:
@benjaminmenetrier I propose to park the Point interfaces for now, if there's no pressing need for them. We can revisit it again at a later point, perhaps with a more native fortran structure that does not need to jump through language interfaces to access the components.
Thanks again!

PR #50, opened 2020-05-13
Custom large grids are too expensive

pmarguinaud:
Structured::crop requires too much memory and time. See the included test case.

This branch attempts to fix this problem for lat/lon grids.

### DIFF
--- a/src/atlas/grid/detail/grid/Structured.cc
+++ b/src/atlas/grid/detail/grid/Structured.cc
@@ -357,10 +357,10 @@ void Structured::crop( const Domain& dom ) {
 
     auto rect_domain = RectangularDomain( dom );
 
-    bool same = true;
+    bool samerows = true;
 
-    for (idx_t j = 0; same && (j < ny ()); j++)
-      same = same && (xmin (j) == xmin (0)) && (dx (j) == dx (0));
+    for (idx_t j = 0; samerows && (j < ny ()); j++)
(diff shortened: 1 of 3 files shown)


pmarguinaud:
@wdeconinck 

Hi Willem,

Travis checks are still running, but I know that my new test "test_largegrid.cc" will not pass on the two OSX machines/configurations. This test is here to prove that creating a 40000x20000 grid does not require more than a few tens of Mb. This works on Linux machines, but on OSX, RSS appears to be 12Gb (at least, this is the value returned by getrusage).
Unfortunately, I do not have access to the OSX environment. What should I do to fix this issue ?

Philippe

wdeconinck:
I have a OSX machine I can try it on. You can park it for now until I see what's going on.

wdeconinck:
Hi @pmarguinaud 
I found the reason being that MacOSX counts the `rss` in bytes, whereas other Linux counts it in kbytes, which explains a factor 1000 difference.  I reimplemented the test with an eckit abstraction for resource usage which takes care of it.

I also improved upon your fix to make it as lightweight for reduced grids.
I noted though that the algorithm to compute Gaussian latitudes is also very memory hungry. Luckily Gaussian latitudes for most popular resolutions are precomputed (N2000, N4000, N8000), and take no time / memory at all.

pmarguinaud:
Very good. Thank you Willem.

PR #54, opened 2020-08-20
Add test to demonstrate a problem in interpolation2D

pmarguinaud:
This was reported by A. Piacentini (CERFACS). The problem seems to come from the fact that the mesh object is modified (some extra nodes are added to it when a halo is requested) by the nodeColumns functionspace, although the mesh object is passed by value to the functionspace ctor.

We can easily avoid this issue by creating two distinct mesh objects, but we would like to know whether this is a bug or not.

I am sorry I have forgotten to mark this request as a draft.

### DIFF
--- a/src/tests/interpolation/test_interpolation_structured2D.cc
+++ b/src/tests/interpolation/test_interpolation_structured2D.cc
@@ -389,6 +389,29 @@ CASE( "test_interpolation_structured for vectors" ) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


wdeconinck:
I believe it is a bug. I will look at it now.

wdeconinck:
The test now passes as of commit d9aa52b merged in develop

PR #56, opened 2020-09-10
Try to change Atlas communicator using eckit setDefaultComm

pmarguinaud:
I would like to change temporarily Atlas communicator; I tried using eckit setDefaultComm, but without success. 

Atlas communicator is a static variable, initialized to eckit default communicator on startup : 

```
inline const Comm& comm() {
    static const Comm& _comm = eckit::mpi::comm();
    return _comm;
}
```

Should I change this to : 

```
inline const Comm& comm() {
    return eckit::mpi::comm();
}
```

Or should I introduce a atlas::mpi::setDefaultComm function ?

### DIFF
--- a/src/atlas/parallel/mpi/mpi.h
+++ b/src/atlas/parallel/mpi/mpi.h
@@ -20,8 +20,7 @@ namespace mpi {
 using Comm = eckit::mpi::Comm;
 
 inline const Comm& comm() {
-    static const Comm& _comm = eckit::mpi::comm();
-    return _comm;
+    return eckit::mpi::comm();
 }
 
 inline idx_t rank() {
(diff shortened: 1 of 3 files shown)


wdeconinck:
Thank you for pointing this out. Indeed that is a bug. My intention was to avoid the calls to `eckit:mpi::comm()` repeatedly, and this is was an unforeseen side-effect. Your suggestion to remove `static` is a good idea.

Also I think you forgot to commit the test that shows the problem.

pmarguinaud:
I have just added the missing test case and removed the static _comm.

Regards

wdeconinck:
Thank you for your contribution!

PR #57, opened 2020-09-22
StructuredColumns with x/y bi-periodicity on regional grids

pmarguinaud:
X periodicity was already working, there was nothing to do. 

The config parameter "periodic_y" enables y-periodicity. Everything works, except that we get spurious coordinates in the halo on domain edges, but this should not be an issue as these coordinates are not defined. I have also added the "periodic_x" config parameter which could be useful in the future.

A test case has been added, which makes some basic checks; I could add some more if necessary.

### DIFF
--- a/src/atlas/functionspace/detail/StructuredColumns_setup.cc
+++ b/src/atlas/functionspace/detail/StructuredColumns_setup.cc
@@ -90,6 +90,10 @@ void StructuredColumns::setup( const grid::Distribution& distribution, const eck
         throw_Exception( "Grid is not a grid::Structured type", Here() );
     }
 
+    bool periodic_x = false, periodic_y = false;
+    config.get ("periodic_x", periodic_x);
+    config.get ("periodic_y", periodic_y);
+
     const double eps = 1.e-12;
 
     ny_                  = grid_->ny();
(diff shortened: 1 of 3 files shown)


src/atlas/functionspace/detail/StructuredColumns_setup.cc
```diff
@@ -310,18 +334,8 @@ void StructuredColumns::setup( const grid::Distribution& distribution, const eck
...
-        ii = compute_i( i, jj );
-        if ( jj != j ) {
-            ATLAS_ASSERT( grid_->nx( jj ) % 2 == 0 );  // assert even number of points
-            ii = ( ii < grid_->nx( jj ) / 2 ) ? ii + grid_->nx( jj ) / 2
-                                              : ( ii >= grid_->nx( jj ) / 2 ) ? ii - grid_->nx( jj ) / 2 : ii;
-        }
-        p = distribution.partition( global_offsets[jj] + ii );
-        return p;
+    auto compute_p = [&compute_g, &distribution]( idx_t i, idx_t j ) -> int {
+        return distribution.partition (compute_g (i, j) - 1);
```
wdeconinck, line 338: Nice simplification

wdeconinck (approved):
Hi Philippe, thanks for your contribution. I think it looks very good. I am not sure what you mean with the spurious domain-edge coordinates however.

pmarguinaud:
Hi Willem,

I mean that the xy field attached to the StructuredColumns object has wrong values for points lying outside the domain. 

For instance, in the test case I provided (80x80 lat/lon grid, 20N..60N, 20E..60E), the point South-West of the South-West corner (20W, 20E) has the following values for x/y : (-0.5063,  60.0000), which is not very nice. Note that all points inside the domain are OK.

I can look at it again if you think it is important.

Philippe

wdeconinck:
Those points would need to have correct coordinates in case of finite-difference stencil computations needing halo points.
If this PR does not change the previously existing behaviour, and this is OK for you at the moment, you could create a follow-on issue to fix this, if you do not have the time right now. Then I merge it in as is.

pmarguinaud:
For me, it is not a problem at all, as I do not intend to use these coordinates. 

Please note that we have two possibilities here for the coordinates of the point SE of the SE corner : 
1- either we extrapolate X/Y coordinates, which looks OK for X, but is impossible for Y, at least for the general case where Y is not necessarily regular
2- or we use the coordinates of the NW point which is the one we have here in the case of X/Y bi-periodicity

I have looked at that this morning, but modifying compute_x_fast interferes with the section of code tagged with ATLAS-186

Note also that method 2 could be easily implemented with a halo exchange on xy.

PR #58, opened 2021-02-19
Distribution setup hangs with L160000x80000

pmarguinaud:
The band distribution setup hangs when distributing a L160000x80000 grid on two tasks.

This branch solves the problem and implements a new test.

Please note that defining a L160000x80000 requires compiling Atlas with -DATLAS_BITS_LOCAL=64.


wdeconinck:
Thanks Philippe! 👍

PR #59, opened 2021-03-01
Bugfix/pointcloud fs

MarekWlasak:
I think that this is a small bugfix needed for using PointCloud function-space.


wdeconinck:
Hello Marek. This is actually a breaking change, as you can see by failing travis tests. 
The "datatype"  value is expected to be the underlying KIND integer. This is in retrospect the wrong choice, and a string value would indeed be better ("real32", "real64", etc.) as it is human readable.
I am happy for this PR to evolve into changing this everywhere, if you're up for it ?

MarekWlasak:
O.k. I partially know what happened .... This work started because
```
 if ( !config.get( "datatype", kind ) ) {
```
was not working as intended. Now it seems to work.  Go figure!

Unfortunately, I don't have the time to do this here and now.  I can only tackle this in a month or two's time. Would you want this to still go through "option"s?  Why do we need options when we have a configuration?

MarekWlasak:
I am happy for this PR to be closed (for now). I have created an issue for the underlying issue.

wdeconinck:
> Unfortunately, I don't have the time to do this here and now. I can only tackle this in a month or two's time. 

As you want.

> Would you want this to still go through "option"s? Why do we need options when we have a configuration?

Going through options::datatype() is just a shorthand for writing util::Config("datatype", int(kind) ), and makes it more easy to show intent and avoid typos.

PR #66, opened 2021-05-24
Cubed sphere grid, projection and mesh generator (based on PR #48_

MarekWlasak:
This PR is an update on PR #48 responding to Willem's comments.


wdeconinck:
I have done some clean up and refactoring which was quicker to just do myself than to just mention it all, while it also gave me the chance to test the code and understand it all.

src/atlas/grid/CubedSphereGrid.h
```diff
@@ -0,0 +1,251 @@
...
+
+#include "atlas/grid/Grid.h"
+#include "atlas/grid/detail/grid/CubedSphere.h"
+
+namespace atlas {
+
+//---------------------------------------------------------------------------------------------------------------------
+class CubedSphereGrid;
+class EquiDistCubedSphereGrid;
+class EquiAnglCubedSphereGrid;
```
wdeconinck, line 20: I'm not seeing where the classes `EquiDistCubedSphereGrid` and `EquiAnglCubedSphereGrid` are implemented. I don't think they have a place either as they don't expose any different API. They could only serve as a convenience constructor:
```
auto grid = EquiDistCubedSphereGrid( 32 );
```
but this could also be achieved already with
```
auto projection = Projection("cubedsphere_equidistant", params);
auto grid = CubedSphereGrid( 32, projection ); // previous line could be inlined here
```

wdeconinck (changes requested):
Thanks for the huge work done by all.

I would still like to see more unit tests for the projections both equiangular and equidistant, with and without stretching factor, that show that we can do round-robin going from
- xy to lonlat to xy
- xy to abt to xy
- ... ?

as well as testing with a few "known" coordinates to verify the projections do what is expected. The projections are now resolution independent so that makes things easier.

I would further like to discuss naming of certain functions and configuration options to be more in line with other Atlas classes. Also to discuss further is how the rotation of the entire grid is described. Atlas already has a rotation projection which is described by setting a new position of the North Pole or South Pole. It would be good to have this streamlined.

Meshgenerator seems to work in "serial". The equal_regions partitioner seems to have an out-of-bounds problem when applied, e.g. with
```
mpirun -np 4 atlas-meshgen CS-EA-32
```
I will investigate this further myself.

Also in a follow-up PR for the mesh generator, it would be cool to (optionally) have the introduction of ghost points on each cube face. This is anyway what a customised cubedsphere partitioner will give us. If we then look at the mesh in e.g. xy coords or latlon coordinates, then we will not have the connecting edges crossing the domain as we see now.

wdeconinck:
Hi @MarekWlasak , you can just continue in this same branch, no problem for me.

wdeconinck:
I know making the mesh generator fully working as desired, in parallel, will be a follow up PR.
But by trying to get the equal-regions partitioner to work with the grid, and to visualise it, I had to modify the mesh-generator a bit.
Also what is apparent is that we need a following function in the CubedSphere:
```
/// return a global index as would be incremented by the iterator
gidx_t CubedSphere::index(idx_t t, idx_t i, idx_t j);
```
This is because the order of points seems different from the iterator when doing:
```
// Following gives wrong global_index of the mesh nodes (and is what is currently used)
idx_t inode{0};
for( idx_t t=0; t<6; ++t ) {
  for( idx_t i=0; i<N; ++i ) {
    for( idx_t j=0; j<N; ++j ) {
        auto p_xy = grid.xy(i,j,t);
        xy(inode,XX) = p_xy.x();
        xy(inode,YY) = p_xy.y();
        part(inode) = distribution.partition(inode); // wrong index used for partitioner
        ++inode;
     }
  }
}
```
Perhaps in the above loop-nest we need to interchange i,j and it would work?
The following ```grid.index(t,i,j)``` function would be more general, and will work in a parallel context where only parts of the grid are visited, as well as in an OpenMP context as no counter like ```inode``` is incremented.
```
// Following would give correct global_index of the mesh nodes
for( idx_t t=0; t<6; ++t ) {
  for( idx_t i=0; i<N; ++i ) {
    for( idx_t j=0; j<N; ++j ) {
        idx_t inode = grid.index(t, i, j);
        auto p_xy = grid.xy(i,j,t);
        xy(inode,XX) = p_xy.x();
        xy(inode,YY) = p_xy.y();
        part(inode) = distribution.partition(inode);
     }
  }
}
```
To work around not having above function ```grid.index(t,i,j)```, I have myself developed a new "t,i,j" iterator for the CubedSphere but not yet committed, it could still be useful to keep perhaps but it is something not available for any other grid yet. In general such iterator should not be required, except in this case, and the function could help.
```
idx_t inode{0}
for( cont auto& p : grid.tij() ) {   // p is of type PointTIJ which is private to CubedSphere
  idx_t t = p.t();
  idx_t i = p.i();
  idx_t j = p.j();
  auto p_xy = grid.xy(t,i,j);
  xy(inode,XX) = p_xy.x();
  xy(inode,YY) = p_xy.y();
  part(inode) = distribution.partition(inode);
  inode++;
}
```

I then managed to correctly assign a partition-number to the nodes/elements which I can visualise with Gmsh. However every MPI-rank outputs the full mesh rather than a single partition because the mesh generator is not yet parallel. I can add all the the modifications I have done to create this after cleaning it up a bit more.

<p align="center">
  
  <b>CS-EA-32 with equal_regions partitioner in 32 parts</b>
</p>

It should be clear that with the grid partitioned in this way, you cannot really take advantage of the structure, but it should be sufficient for certain interpolation algorithms etc.

I noticed that the grid iteration is quite heavy with a lot of if-statements with value comparisons. Not really my personal concern but there's a lot of room for optimisations, using some counters and small arrays of offsets etc.

src/atlas/projection/detail/CubedSphereProjectionBase.cc
```diff
@@ -327,9 +327,27 @@ void CubedSphereProjectionBase::alphabetat2xy(const idx_t& t, const double ab[],
...
+    if (is_same(xy[XX], 0.0, tol)) { xy[XX] = 0.0; }
+    if (is_same(xy[XX], 90.0, tol)) { xy[XX] = 90.0; }
+    if (is_same(xy[XX], 180.0, tol)) { xy[XX] = 180.0; }
+    if (is_same(xy[XX], 270.0, tol)) { xy[XX] = 270.0; }
+    if (is_same(xy[YY], -45.0, tol) && (xy[XX] <= 180.0)) { xy[YY] = -45.0; }
+    if (is_same(xy[YY], 45.0, tol) && (xy[XX] > 180.0)) { xy[YY] = 45.0; }
+    if (is_same(xy[YY], 45.0, tol) && is_same(xy[XX], 0.0)) { xy[YY] = 45.0; }
+    if (xy[YY] >= 135.0) { xy[YY] = 135.0 - __DBL_EPSILON__ ; }
+    if (xy[YY] <= -135.0) { xy[YY] = -135.0 + __DBL_EPSILON__ ; }
+    if (xy[XX] >= 360.0) { xy[XX] = 360.0 + __DBL_EPSILON__ ; }
```
wdeconinck, line 349: Why do you need to add/subtract ```__DBL_EPSILON__``` , and not just snap to value as in
```
xy[YY] = std::min( xy[YY],  135.);
xy[YY] = std::max( xy[YY], -135.);
xy[XX] = std::min( xy[XX],  360.);
xy[XX] = std::max( xy[XX],    0.);
```
You should also use ```std::numeric_limits<double>::epsilon()``` defined in header ```<limits>``` which is equivalent to ```__DBL_EPSILON__```

MarekWlasak:
@wdeconinck - Thank you for the work you have done on the mesh partitioning.  I have rewritten some of the tests.  I do have a question though.  What is the best way to explicitly test protected members of CubedSphereProjectionBase? Should they be moved  into the ProjectionUtilities Class? ... Or should there be a CubedSphereProjectionUtilities class which they should go in?

src/tests/grid/test_grids.cc
```diff
@@ -427,55 +427,70 @@ CASE( "test_structured_from_config" ) {
...
+            grid->projection().lonlat2xy(crd);
+            pointXYs_from_LonLat.push_back(crd);
+        }
+        EXPECT(pointLonLats.size() == 6* resolution * resolution +2);
+        EXPECT(pointXYs.size() == 6* resolution * resolution +2);
+        EXPECT(grid.size() == 6* resolution * resolution +2);
+
+        // Note that with nodal points on the cubed-sphere
+        if (resolution == 2) {
+            const double rpi     = 2.0 * asin( 1.0 );
```
wdeconinck, line 459: You can use
```
constexpr double rpi = M_PI;
constexpr double rad2deg = 180. / rpi;
```
where ```M_PI``` is defined in header ```<cmath>```
Also use ```std::asin``` and ```std::atan```

src/atlas/projection/detail/CubedSphereProjectionBase.cc
MarekWlasak: To be well defined I need values in xy to be less than 135 degree in Y more than -135 in Y and less than 360 in X.  For that I need epsilon.  I will rewrite this with min / max.

wdeconinck:
@MarekWlasak which functions would you like to test that are private/protected? It smells to me that you want to test something that deserves to be a class or function on its own then, which you can then compose into the CubedSphereProjectionBase class as member.

MarekWlasak:
@wdeconinck - You previously suggested creating tests for alphabetat2xy and the inverse. I was trying to look at that. Do you think they should be tested separately? There are already tested implicitly when calculating xy and lonlat.

src/tests/grid/test_grids.cc
MarekWlasak: Done

wdeconinck:
You are right that they are tested implicitly. Good enough for me.
Once there's a problem (this test fails), you can narrow it down if those functions are the problem.

wdeconinck:
@MarekWlasak I have fixed a "FE_DIVBYZERO" signal with 94e83f4 which happened on one of the travis builds (PGI 20.9), but exposes now a failed "EXPECT":
```
[0] EXPECT_APPROX_EQ( lonlat, crd, 1e-11 ) FAILED @ test_projection_cubed_sphere.cc +38
[0]  --> lhs = {90,82.5}
[0]  --> rhs = {89.999999146226,82.500000000000}
```
See here for the full log:  https://travis-ci.org/github/ecmwf/atlas/jobs/772646097#L1620
If you could try to find a solution for this, would be great. Problem seems to be it is only this build which is problematic.

MarekWlasak:
@wdeconnick - I have chased the error/difference from the other builds for this point. It seems to be coming from Spherical2Cartesian calculation. 
```
xy2lonlat start xy = 135 82.5
sphericalToCartesian:: pointLonLat pointXYZ = {0,-7.500000000000009} {0.9914448613738104,0,-0.1305261922200517}
end of equiangular xy2lonlat lonlat = 89.99999999999997 82.5
equiangular lonlat2xy start : lonlat = 89.99999999999997 82.5
sphericalToCartesian:: pointLonLat pointXYZ = {89.99999999999997,82.5} {0,0.1305261922200518,0.9914448613738104}
tileFromLonLat:: lonlat abs xyz t = -0.9914448613738104 -0.8609186691537586 -0.9914448613738104 -1.121971053593862 1.570796326794896 1.439896632895322 0 0.1305261922200518 -0.9914448613738104 2
sphericalToCartesian:: pointLonLat pointXYZ = {89.99999999999997,82.5} {0,0.1305261922200518,0.9914448613738104}
equiangular lonlat2xy xyz ab : 0.9914448613738104 6.070848880062641e-17 0.1305261922200517 6.123233995736766e-17 -0.1308996938995748
enforcXYDomain before 135 82.5
enforcXYDomain after 135 82.5
equiangular lonlat2xy end : xy = 135 82.5
3xy2lonlat start xy = 135 82.5
```
In this case X coordinate is 0 while in the failing test it is -1.9e-9.






```

File: src/atlas/grid/detail/grid/CubedSphere.cc [hunk 4]
```diff
@@ -0,0 +1,355 @@
...
+    double normalisedY = (xy[YY] + 135.)/90.;
+
+    double CubeNxDouble = static_cast<double>(CubeNx_);
+
+    std::vector<double> yOffset{CubeNxDouble,
+                                CubeNxDouble,
+                                2. *  CubeNxDouble,
+                                CubeNxDouble,
+                                CubeNxDouble,
+                                0};
```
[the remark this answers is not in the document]
MarekWlasak: Done!

src/atlas/grid/detail/grid/CubedSphere.cc
```diff
@@ -0,0 +1,355 @@
...
+        dynamic_cast<const CubedSphereProjectionBase &>(projection_).tileFromXY(xy);
+}
+
+// Convert from xyt space into continuous xy space.
+void CubedSphere::xyt2xy(const double xyt[], double xy[]) const {
+    // xy is in degrees
+    // while xyt is in number of grid points
+    // (alpha, beta) and tiles.
+    std::vector<double> xOffsetDeg{0., 90., 90., 180., 270., 270.};
+    std::vector<double> yOffsetDeg{-45., -45., 45., -45., -45., -135.};
```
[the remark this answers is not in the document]
MarekWlasak: Done!

src/atlas/grid/detail/grid/CubedSphere.cc
```diff
@@ -0,0 +1,355 @@
...
+
+    double N = static_cast<double>(CubeNx_);
+    std::vector<double> xOffsetIndex{0, N, N, 2*N, 3*N,  3*N};
+    std::vector<double> yOffsetIndex{N, N, 2*N, N,  N, 0};
```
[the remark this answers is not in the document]
MarekWlasak: Done!

src/atlas/grid/CubedSphereGrid.h
MarekWlasak: Removed the other classes

MarekWlasak:
@wdeconinck - I have dealt with the comments in the ticket. There is still the problem with the single point (90,82.5) which gives a relative error much larger than it should, but that is with only one compiler and it seems to stem from spherical2cartesian transformation.  I have reduced the tolerance of the test to 1e-6 for now, so that all tests pass.  In a future PR it might be worth to creating a separate test for spherical2cartesian as back.  Is that o.k?

MarekWlasak:
@wdeconinck - What more do I need to do to get this PR in?

wdeconinck:
> @wdeconinck - I have dealt with the comments in the ticket. There is still the problem with the single point (90,82.5) which gives a relative error much larger than it should, but that is with only one compiler and it seems to stem from spherical2cartesian transformation. I have reduced the tolerance of the test to 1e-6 for now, so that all tests pass. In a future PR it might be worth to creating a separate test for spherical2cartesian as back. Is that o.k?

That is OK

I will test this on our supercomputer as well and if succesful, merge it to develop.

wdeconinck (approved):

wdeconinck:
Thanks all, @danholdaway , @MarekWlasak , @odlomax for this nice contribution!

PR #67, opened 2021-06-22
Feature/cubed sphere tiles

MarekWlasak:
This is part 1 of the changes needed to allow for the LFRic tile strategy.  

Part 2 will look at the global indexing in the CubedSphere grid class.

My main concern with this approach is that it might be slower than what is currently there


src/atlas/grid/Tiles.h
```diff
@@ -0,0 +1,129 @@
...
+      std::cout << "grid/Tiles.h LFRicCubedSphereTiles"  << std::endl;
+    }
+
+    LFRicCubedSphereTiles( const CubedSphereTiles& );
+
+private:
+    const ::atlas::cubedspheretiles::LFRicCubedSphereTiles* cubedspheretiles_;
+};
+
+
```
wdeconinck, line 128: The classes `atlas::LFricCubedSphereTiles` and `atlas::FV3CubedSphereTiles` bring nothing extra in terms of API from `atlas::CubedSphereTiles` besides perhaps the empty constructor.
I would like to have the implementations to be hidden more, so that it would be enough to just use
```
auto tiles = atlas::CubedSphereTiles("LFric");
```
and/or via an `eckit::Parametrisation`
```
auto tiles = atlas::CubedSphereTiles( atlas::options::type("LFRic") );    // atlas::util::Config argument
```
The first constructor using a type string could just call the second one using Parametrisation.

src/atlas/grid/detail/tiles/LFRicTiles.h
```diff
@@ -0,0 +1,73 @@
...
+
+    virtual void tile3RotateInverse( double xyz[] ) const override;
+
+    virtual void tile4RotateInverse( double xyz[] ) const override;
+
+    virtual void tile5RotateInverse( double xyz[] ) const override;
+
+    virtual idx_t tileFromXY( const double xy[] ) const override;
+
+    virtual idx_t tileFromLonLat( const double lonlat[] ) const override;
```
wdeconinck, line 61: It is expected in Atlas that a lonlat is given in degrees. Please modify implementation and use case to use degrees instead.

FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/atlas?pullRequest=67) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/atlas?pullRequest=67) before we can accept your contribution.<br/>**2** out of **3** committers have signed the CLA.<br/><br/>:white_check_mark: odlomax<br/>:white_check_mark: MarekWlasak<br/>:x: KrissyRayD242<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/atlas?pullRequest=67) it.</sub>

src/atlas/grid/detail/tiles/LFRicTiles.h
MarekWlasak: Done!

src/atlas/grid/Tiles.h
MarekWlasak: Done - Handles now removed - I can now do 
```
auto tiles = atlas::CubedSphereTiles("LFRicCubedSphereTiles");
```
but not the latter.

MarekWlasak:
@wdeconinck - I have implemented your suggestions and have done part 2 i.e. the part involving the grid class.

src/atlas/grid/Tiles.h
wdeconinck: I noticed that you use "tile type" as configuration. It would be better to just use "type", which is more in line with Atlas, so that a configuration would be
```
grid:
    type: cubedsphere
    N : 32
    tiles: { type: LFRicCubedSphereTiles }
    projection: { type: cubedsphere_equiangular }
    ... ? ...
```
Which brings me also that perhaps for "nice" reading the name of `LFricCubedSphereTiles::static_type()` could just be "cubedsphere_LFRic" (or lowercase) for consistency also with the projections:
```
grid:
    type: cubedsphere
    N : 32
    tiles:      { type: cubedsphere_lfric }
    projection: { type: cubedsphere_equiangular }
    ... ? ...
```

MarekWlasak: @wdeconinck - I can easily do the cosmetic change changing "LFRicCubedSphereTiles" to cubedsphere_lfric. However, I have a problem in that I need the methods that tiles class gives in the CubedSphereProjectionBase class, so that they can be used the equiangular and equidistant projections.  If it was not for the projection class I could add a Tiles class argument to the constructor of CubedSphere class.  Is there a way of getting a subconfiguration from a parameter class?

i.e.
```
grid:
    type: cubedsphere
    N : 32
    projection:
        type: cubedsphere_equiangular
       ...
        tiles:
           type: cubedsphere_lfric
    ... ? ...
```

MarekWlasak:
@wdeconinck - Hi ! I have had a chat with Wojciech and I think I can partially solve this issue where internally to the Tile class it uses "type", but externally when constructed from the projection class it has "tile type" that gets copied into "type" for tile class.  Is that good enough?

wdeconinck:
> @wdeconinck - Hi ! I have had a chat with Wojciech and I think I can partially solve this issue where internally to the Tile class it uses "type", but externally when constructed from the projection class it has "tile type" that gets copied into "type" for tile class. Is that good enough?

I see the problem. You can use following then:
```
grid:
  type: cubedsphere
  N: 32
  projection:
    type: cubedsphere_equiangular
    tiles: { type : cubedsphere_lfric }
```

Upon construction of the `CubedSphere` grid, you could access the tiles from the projection:
```
class CubedSphere {
...
private:
  CubedSphereProjectionBase* cs_projection_;  // store pointer to dynamic_cast for convenience
  Tiles tiles_;                               // retrieve from projection during constructor
};
CubedSphere::CubedSphere(...) { 
  ...
  cs_projection_ = dynamic_cast<CubedSphereProjectionBase*>( projection().get() );
  tiles_ = cs_projection_->tiles();
}
```

Another neat trick to set (or get) sub configurations programmatically, is using the "."
```
atlas::util::Config grid;
grid.set("type","cubedsphere");
grid.set("N",32);
grid.set("projection.type","cubedsphere_equiangular");
grid.set("projection.tiles.type,"cubedsphere_lfric");
```

MarekWlasak:
@wdeconinck -  I think that I have implemented what you wanted.

MarekWlasak:
@wdeconinck - I think I have done all that you have asked.

wdeconinck:
Hi @MarekWlasak, that looks all good to me now. Great work.
As you noticed Travis has stopped working. This needs to be fixed when I have time. Probably we will move to Github Actions instead.
I will take the branch for a spin today, and test on all our available systems, and probably apply clang-format before merging.

MarekWlasak:
@wdeconinck - when do you think that you will have time to put this PR in? We have about 4 other PRs that we could send you ie.
  *  cell centered grid
  *  cube periodicity
  *  cubed-sphere partitioner
  *  cell-centered cubed-sphere mesh generator.
 but we don't want to overload you.

wdeconinck:
@MarekWlasak sorry I was side-tracked with something urgent and some holidays. I will have time tomorrow.

MarekWlasak:
@wdeconinck Great! Thank you!

wdeconinck:
I would have thought this PR would take care of following `if / else if` statement:
https://github.com/ecmwf/atlas/blob/fd616c11cf3beb60aaf8d2bb583d92d31041003f/src/atlas/grid/detail/grid/CubedSphere.cc#L95
and matching
https://github.com/ecmwf/atlas/blob/fd616c11cf3beb60aaf8d2bb583d92d31041003f/src/atlas/grid/detail/grid/CubedSphere.cc#L138

MarekWlasak:
@wdeconinck - I have created an issue to ensure that the above point does not get lost!

wdeconinck:
@MarekWlasak I reviewed the code and merged it into the develop branch.
I made some cosmetic changes, and some changes related to tolerances that made some of our platforms fail.

PR #70, opened 2021-08-25
CubedSphere - partitioner, cubed-sphere periodicity function and other tweaks

MarekWlasak:


wdeconinck:
Thank you @odlomax and @MarekWlasak this looks great.
I have made a few fixes required to compile, and also configured the default meshgenerator and partitioner depending on the staggering.

I noticed a few things that don't work still:
- lfric tiling with 1 element per tile:
  ```atlas-meshgen CS-LFR-C-1``` --> FE_DIV_BY_ZERO exception rather than nice message
- fv3 tiling with centroid grid:
  ```atlas-meshgen CS-EA-C-10```
   --> "Assertion failed: nodeGlobalIdxCount == nNodesUnique + 1 in generate_mesh, line 441 of CubedSphereMeshGenerator.cc"
- lfric tiling with nodal grid:
  ```atlas-meshgen CS-LFR-L-10``` --> seems to run without mistake, but mesh is wrong when visualised with Gmsh
- The cubedsphere partitioner only works for centroid grids:
   ```mpirun -np 24 bin/atlas-meshgen CS-LFR-L-10 --partitioner=cubedsphere``` 
    --> "Assertion failed:  [[ part[nb_cells - 1] == nb_partitions() - 1 ]]
number of partitions created not what is expected in partition, line 224 of /lus/pfs1/hpcperm/nawd/atlas-bundle/source/atlas/src/atlas/grid/detail/partitioner/CubedSpherePartitioner.cc"

Please fix if these cases are straightforward to fix, otherwise please make sure we get a "NotImplemented" exception where appropriate. Especially the silent failure (bullet 3) is dangerous.

wdeconinck:
Added one more commit where now names like "CS-LFR-16" refer to "CS-LFR-C-16"
I am inclined to think that in the future we will have a different name for the "CS-LFR-L-16" grid, or it will not be required at all.

wdeconinck:
The PR has been squash-merged in develop with commit  b1f8ff0
Above mentioned issues will be fixed in upcoming PR by @odlomax .

PR #72, opened 2021-09-14
Feature/interpolation adjoint

MarekWlasak:
This implements a capability to do the adjoint of the interpolation in atlas where sparse matrices are used.
I have tested this with bi-linear on structured grid.


MarekWlasak:
I have now also tested it with MPI (4 pes) and now can do the interpolation from Gaussian F* to regular lon lat grids and adjoint. L* and S*.

wdeconinck:
I added a few minor commits to perfect it. I will now test on our CI machines and then merge when successful

wdeconinck:
Squash-merged with commit  6a45cab

PR #73, opened 2021-09-14
Cubed sphere parallel mesh generator with halos.

odlomax:
### This is an updated version of [this PR](https://github.com/JCSDA-internal/atlas/pull/58), that should be mergeable with the ecmwf/develop branch. Original text is as follows:

This PR implements the multi-PE cubed-sphere mesh generator with halos.

The mesh generator internally treats the tiles of the cubed-sphere as separate overlapping meshes. Halos are added at tile boundaries and at partition boundaries on tile interiors.

In addition to the usual `mesh.cells()` and `mesh.nodes()` fields, the following the fields are added to mesh cells/nodes objects:

* **xy**. The xy positions of owned data follow the conventions give in the `Tiles` class. The xy projection of halo data is contiguous with the formal xy projection of the owned data. This should will make interpolation significantly simpler. Cells and nodes both have xy fields.

* **lonlat**. Cells receive lonlat coordinates as well as nodes.

* **ijt**. Cells and nodes have fields giving the (i, j, t) coordinates of the data. Halo (i, j, t) indices are extrapolated from the owned data, i.e. i and j can be less than zero and greater than N for tile-boundary halos. This information can be used to construct a structured `CubedSphere` function space, similar to `StructuredColumns`.

Utility classes were written for the mesh generator in `src/atlas/meshgenerator/detail/cubedsphere/CubedSphereUtility.h, CubedSphereUtility.cc`. The main class is `NeighbourJacobian`. This handles the xy and ij coordinate transforms between a tile and its four neighbours. I've put it here for the time being, but there's probably a better home for it. It's indispensable for finding the ij and xy positions of halos. I also believe it will be useful for the interpolation of vector fields.

_Fig. 1. The xy projection for an N12 cubed sphere mesh with `cubed_sphere` partitioner and `halo = 3` on 8 PEs. Some of the halos overlap in this projection and are not visible in the figure._
allhalo3xyz

_Fig 2. The mesh on PE 7 from Fig. 1. Note that the xy projection of the halos is contiguous with owned data projection._
part7halo3xy

_Fig. 3. As Fig. 2, but projected into xyz space using the lonlat coordinates. Note that "gaps" in the cross shaped halo distribution in Fig. 2 do not exist in 3D space._
part7halo3xyz

I've added test executable `src/tests/mesh/test_cubedsphere_meshgen.cc`

This performs two tests.
* The first test makes sure the `NeighbourJacobian` can correctly place halo points around an `N2` grid.
* The second test runs the cubed sphere mesh generator with different partitioner and halo-size configurations. It ensures that a simple test field (see below) can be written to owned nodes/cells and distributed to halo nodes cells via the Atlas halo-exchange methods.

_Fig 4. A simple test field with f(lon, lat) = sin(3 * lon) * sin(2 * lat)_
testfield

### Changes since last PR.

In addition to the above, I've refactored the code so that it uses significantly less memory than before. It still has to construct _some_  global information, but I've limited it to 24 bytes for each global node and each global cell. Further memory is allocated for local information, but this is restricted to _only_ the nodes/cells on the PE. It can build an N1280 mesh (~10^7 elements) using somewhere around 500MB per PE.

I've also improved the accuracy of the mesh generator and tested this with the horrible grid size N211 (a prime number, yuck!). The halo nodes/cells and their respective owned nodes/cells are now in the same position to a precision near machine epsilon.

Finally, I've added logic to properly number and sort the halo fields. A CellColumns and NodeColumns halo field from an N12 grid with equal_regions partitioning are shown below:

equal_regions_cell_halo
equal_regions_node_halo

### DIFF
--- a/src/atlas/CMakeLists.txt
+++ b/src/atlas/CMakeLists.txt
@@ -381,6 +381,8 @@ meshgenerator/detail/MeshGeneratorInterface.cc
 meshgenerator/detail/MeshGeneratorInterface.h
 meshgenerator/detail/HealpixMeshGenerator.h
 meshgenerator/detail/HealpixMeshGenerator.cc
+meshgenerator/detail/cubedsphere/CubedSphereUtility.h
+meshgenerator/detail/cubedsphere/CubedSphereUtility.cc
 )
 
 list( APPEND atlas_output_srcs
--- a/src/atlas/grid/detail/tiles/FV3Tiles.cc
+++ b/src/atlas/grid/detail/tiles/FV3Tiles.cc
@@ -444,8 +444,7 @@ void FV3CubedSphereTiles::enforceXYdomain( double xy[] ) const {
 
 atlas::PointXY FV3CubedSphereTiles::tileCubePeriodicity( const atlas::PointXY& xyExtended,
                                                          const atlas::idx_t tile ) const {
-    atlas::PointXY xyOwned;
-    return xyOwned;
+    throw_NotImplemented("tileCubePeriodicity not implemented for FV3Tiles", Here());
 }
 
 void FV3CubedSphereTiles::print( std::ostream& os ) const {
--- a/src/atlas/grid/detail/tiles/FV3Tiles.cc
+++ b/src/atlas/grid/detail/tiles/FV3Tiles.cc
@@ -444,7 +444,7 @@ void FV3CubedSphereTiles::enforceXYdomain( double xy[] ) const {
 
 atlas::PointXY FV3CubedSphereTiles::tileCubePeriodicity( const atlas::PointXY& xyExtended,
                                                          const atlas::idx_t tile ) const {
-    throw_NotImplemented("tileCubePeriodicity not implemented for FV3Tiles", Here());
+    throw_NotImplemented( "tileCubePeriodicity not implemented for FV3Tiles", Here() );
 }
 
 void FV3CubedSphereTiles::print( std::ostream& os ) const {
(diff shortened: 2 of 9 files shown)


wdeconinck:
Hi @odlomax that is looking great! I am testing it and will merge it today.

odlomax:
> Hi @odlomax that is looking great! I am testing it and will merge it today.

Thanks, Willem. @svahl991 spotted some typos (`and` vs `&&` and the like!). Is it okay if I push a correction for them?

wdeconinck:
> > Hi @odlomax that is looking great! I am testing it and will merge it today.
> 
> Thanks, Willem. @svahl991 spotted some typos (`and` vs `&&` and the like!). Is it okay if I push a correction for them?

Yes sure, go ahead.

wdeconinck:
@odlomax Eigen is a optional dependency at the moment and you are using Eigen directly in the CubedSphereUtility  Jacobian2 class.
We have a Matrix in eckit which is implemented either by Eigen directly, or by a "own" implementation "MatrixLapack"
See https://github.com/ecmwf/eckit/blob/develop/src/eckit/maths/Matrix.h
Please try compiling eckit and atlas with "-DENABLE_EIGEN=OFF" to see if you could find a working solution.

wdeconinck (changes requested):
Eigen should remain optional, at least for now. Some compilers like Cray cce/8.7 can not deal with certain simple Eigen operators.

odlomax:
> @odlomax Eigen is a optional dependency at the moment and you are using Eigen directly in the CubedSphereUtility Jacobian2 class.
> We have a Matrix in eckit which is implemented either by Eigen directly, or by a "own" implementation "MatrixLapack"
> See https://github.com/ecmwf/eckit/blob/develop/src/eckit/maths/Matrix.h
> Please try compiling eckit and atlas with "-DENABLE_EIGEN=OFF" to see if you could find a working solution.

I've now hand written all the small matrix operations. That gets rid of the Eigen dependency. It's also made the mesh generator faster. Odd, I'd have thought Eigen would be well-optimised for small statically sized matrices. Oh well, can't complain!

wdeconinck:
Squash-merged branch with commit 53d1906

PR #74, opened 2021-09-17
Refactor of adjoint interpolation interface.

MarekWlasak:
Changes are as follows:
1) setting target to const in the adjoint 
2) using ArrayT for the temporary array
3) using .assign(0.) to fill array
4) setting tolerances in tests to have a relative error


wdeconinck (approved):
Looks all good and as we discussed offline.

wdeconinck:
PR is merged in develop with commit 51edfcb

PR #76, opened 2021-09-24
small change to allow StructuredColumns gaussian field to have a halo…

MarekWlasak:
… when using trans.


wdeconinck (changes requested):
Few remarks: it would be better to have the loop bound variables be `const idx_t` before any `for` loop starts, especially when it's not a simple inline function that's being called.

Also it could be possible the field doesn't have the `functionspace()` assigned per se
Best would be to pass the StructuredColumns function space as extra argument besides the Field.

wdeconinck (approved):
Perfect

wdeconinck:
Merged with commit 8949d2b

PR #80, opened 2021-10-21
Feature/redistribution

odlomax:
This is a copy of [this PR](https://github.com/JCSDA-internal/atlas/pull/72), which implements a generic redistribution class.


wdeconinck:
Commit 91ec362 in develop is a squash of this PR. Thanks @odlomax !

PR #82, opened 2021-12-01
Feature/ecmwf ukv

MO-marcomilan:
Providing a stretch projection capability from a LAM with regular structured grid.

Originally this PR was the #73 
https://github.com/JCSDA-internal/atlas/pull/73

I created a new branch from feature/ecmwf-develop in JCSDA internal

The theory used in this code can be found under:
The benefits of the Met Office variable resolution NWP model for forecasting convection, Tang et al. 2013, https://doi.org/10.1002/met.1300

In the system we have the capability of a rim zone


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/atlas?pullRequest=82) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/atlas?pullRequest=82) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/atlas?pullRequest=82) it.</sub>

MO-marcomilan:
ORIGINAL regular grid LAT
xy_lat

ORIGINAL regular grid LON
xy_lon

MO-marcomilan:
Larger regular grid LAT
regular_lat
Larger regular grid LON
regular_lon

MO-marcomilan:
Stretched grid LAT
xy_lat
Stretched grid LON
xy_lon

src/atlas/projection/detail/StretchRegular.h
```diff
@@ -0,0 +1,78 @@
...
+
+template <typename Rotation>
+class StretchLAM final : public ProjectionImpl {
+public:
+    using Spec = ProjectionImpl::Spec;
+
+    ///< constructor uses parametrisation and point to stretch
+    StretchLAM(const eckit::Parametrisation& );
+    ///< projection name
+    static std::string static_type() { return Rotation::typePrefix() + "stretch"; }
```
wdeconinck, line 26: We will need something smarter than "stretch"
Perhaps something like a "dual_resolution" ?

src/atlas/projection/detail/StretchRegular.h
```diff
@@ -0,0 +1,78 @@
...
+    double x_reg_start_; ///< xstart of the internal regional grid
+    double y_reg_start_; ///< ystart of the internal regional grid
+    double x_reg_end_; ///< xend of the internal regional grid
+    double y_reg_end_; ///< yend of the internal regional grid
+    double startx_; ///< original domain startx
+    double endx_; ///< original domain endx
+    double starty_; ///< original domain starty
+    double endy_; ///< original domain endy
+    double rim_widthx_; ///< xsize of the rim
+    double rim_widthy_; ///< ysize of the rim
```
wdeconinck, line 65: What does the rim correspond to?
Is it something like `x_reg_start - startx` ?
Or is it only the width of the transition region ?

src/atlas/projection/detail/StretchRegular.cc
```diff
@@ -0,0 +1,385 @@
...
+    proj_st.get( "rim_widthx", rim_widthx_ ); ///< xsize of the rim
+    proj_st.get( "rim_widthy", rim_widthy_ ); ///< ysize of the rim
+}
+
+template<typename Rotation>
+void StretchLAM<Rotation>::checkvalue(const double & epsilon,
+                                        const double & value_check) const {
+    std::string err_message;
+    std::string str = std::to_string(value_check);
+    err_message = "USER defined limits not in the middle of the area " + str;
```
wdeconinck, line 96: These strings are created regardless of the error. You will see a big speedup if these are instead within the following `if` statement.

src/atlas/projection/detail/StretchRegular.cc
```diff
@@ -0,0 +1,385 @@
...
+              double check_y = starty_ + add_yf_ - phi_start;
+              double check_st = n_stretchedx_ - n_stretchedy_;
+              checkvalue(epsilon, check_x);
+              checkvalue(epsilon, check_y);
+              checkvalue(epsilon, check_st);
+
+         }
+
+    int nx_ = ((deltax_all + epstest) / delta_high_) + 1;
+    int ny_ = ((deltay_all + epstest) / delta_high_ ) + 1;
```
wdeconinck, line 354: The selection is identical for each point that will be projected.
It will be more performant to add this into the constructor instead.

src/atlas/projection/detail/StretchRegular.h
MO-marcomilan: Sorry, if I wasn't clear. When a stretched grid is applied in LAM, the desired grid types/resolutions from outside to centre are:
- Low resolution regular grid (rim). This could have the same resolution of the global model.
- Stretched grid. Variable resolution.
- Internal high resolution regular grid.

src/atlas/projection/detail/StretchRegular.h
```diff
@@ -0,0 +1,78 @@
...
+    std::string type() const override { return static_type(); }
+
+    ///< projection and inverse projection
+    void xy2lonlat( double crd[] ) const override;
+    void lonlat2xy( double crd[] ) const override;
+
+    ///< specification for stretching
+    Spec spec() const override;
+
+    std::string units() const override { return "meters"; }
```
wdeconinck, line 36: The units should be "degrees" if I follow the code correctly.

src/atlas/projection/detail/StretchRegular.cc
MO-marcomilan: Done.

src/tests/grid/test_stretchedRegular.cc
```diff
@@ -0,0 +1,228 @@
...
+    for (size_t i = 0; i < delta.size(); ++i) {
+        delta[i] = v[i + 1] - v[i];
+    }
+    Log::info() << "delta = " << delta << std::endl;
+    // Outputs:
+    // delta = [1,1,0.883845,0.798259,0.720961,0.651148,0.651148,0.651148,0.651148,
+    //          0.651148,0.651148,0.651148,0.651148,0.651148,0.651148,0.651148,0.651148,
+    //          0.651148,0.651148,0.651148,0.651148,0.651148,0.651148,0.651148,0.651148,
+    //          0.651148,0.651148,0.651148,0.720961,0.798259,0.883845,1,1]
+}
```
wdeconinck, line 114: Hi @MO-marcomilan I added this "CASE" in the test must to understand a bit the progression.
As I understand from the configuration, the "var_ratio = 1.13" which means that in the transition zone the delta variables should progress with a factor 1.13, but when looking at the values it is not the case.
Could you elaborate?

wdeconinck:
@MarekWlasak , @MO-marcomilan , to fully understand how one would use this, I would like to see how one could create a grid symmetric centred around a value `{x0,y0}`
with a `delta_low = 1 ` and a `n_low = 11` so that the "inner regular grid" would be between {x0-5,x0+5},{y0-5,y0+5}.
Then make the progression with a factor 1.2, until the value `delta_high = 5` is reached.
I assume we would need 9 steps to reach it `1.2^9 = 5.15978` and `1.2^8 = 4.2998`
Does that mean that we need to "progress" only 8 steps and then the next increment is `5`?
When one is then continuing again with the outer quasi-regular frame, you can only arrive to the full grid bounding box by working your way from inside out? Or how is one to embed it in a predefined 5-degree coarser global grid for instance.

When defining the {x,y} - coordinates space (unprojected), would it make sense to assume it follows the increments of the inner-grid? 
That means that the regular_grid "dx" would then correspond to "delta_high", the real domain of interest.

wdeconinck:
Ideally I would like to see some kind of easy grid builder with very few parameters. For instance I think the "rim_width" could be computed from other parameters, or vice-versa, the outer bounding box can be computed from the rim_width and the inner grid parameters?

src/atlas/projection/detail/StretchRegular.cc
wdeconinck: Thanks, I don't think it is pushed yet. Note that I added a few commits to the branch.

src/atlas/projection/detail/StretchRegular.cc
MO-marcomilan: Hi Willelm,
Thank you for your comments.
The code is still not pushed. This is a general question.
I put the code under the header for the function xy2lonlat. The issue is that the function itself is defined as virtual const, I set the variables I need as mutable. It works but I have a second problem.
Hi all, I changed the header of a file in atlas from:

in the header I have
void xy2lonlat( double crd[]) const override{
mycode
};

In the cpp file I have:
void StretchLAM<Rotation>::xy2lonlat( double crd[] ) const {
some other code
};

the problem is that now I have the error:
"error: redefinition of 'void atlas::projection::detail::StretchLAM<Rotation>::xy2lonlat(double*) const'"

Do you have any idea how to solve this?

Has somebody an idea how to avoid it?

wdeconinck: In the header you only need the signature of the function:
```c++
void xy2lonlat(double crd[]) const override;
```
In the .cc file comes the implementation with all the code:
```c++
void StretchLAM::xy2lonlat(double crd[]) const {
   <the implementation>
}
The function is correctly marked `const` and using mutable member variables is a hack, which is dangerous in e.g. an OpenMP context.
You have to initialize these values in the constructor instead, which I think you can because they are independent of the `crd` argument.

Comment (MO-marcomilan) [reply src/atlas/projection/detail/StretchRegular.cc:354]: I implemented the code in the wrong construnctor. Sorry.
Now is in the file.cc for StretchLAM

File: src/atlas/projection/detail/StretchRegular.h (hunk 5 above)
Comment (MO-marcomilan) [reply src/atlas/projection/detail/StretchRegular.h:36]: done

File: src/atlas/projection/detail/StretchRegular.cc (hunk 4 above)
Comment (MO-marcomilan) [reply src/atlas/projection/detail/StretchRegular.cc]: I also didn't need the use of mutable member, I pushed now my changes.
Thanks

File: src/atlas/projection/detail/StretchRegular.cc [hunk 7]
```diff
@@ -86,17 +87,82 @@ StretchLAM<Rotation>::StretchLAM(const eckit::Parametrisation& proj_st) :
...
 void StretchLAM<Rotation>::checkvalue(const double & epsilon,
                                         const double & value_check) const {
     std::string err_message;
     std::string str = std::to_string(value_check);
-    err_message = "USER defined limits not in the middle of the area " + str;
+    //err_message = "USER defined limits not in the middle of the area " + str;
     if (value_check > epsilon || value_check < (-1. * epsilon))
           {
-             throw eckit::BadValue(err_message, Here());
+             throw eckit::BadValue("USER defined limits not in the middle of the area " + str, Here());
```
wdeconinck, line 165: Already better, but you can also bring 
```c++
std::string err_message;
std::string str = std::to_string(value_check);
```
inside the `if` although now it is not as critical anymore since it is only done a few times in the constructor only rather than for each `xy2lonlat()`

src/atlas/projection/detail/StretchRegular.h
```diff
@@ -27,13 +29,16 @@ class StretchLAM final : public ProjectionImpl {
     std::string type() const override { return static_type(); }
 
     ///< projection and inverse projection
-    void xy2lonlat( double crd[] ) const override;
+    /// //void xy2lonlat( double crd[] ) const override;
```
wdeconinck, line 32: Remove extra line
```suggestion
```

src/atlas/projection/detail/StretchRegular.h
```diff
@@ -27,13 +29,16 @@ class StretchLAM final : public ProjectionImpl {
...
+
+    void xy2lonlat( double crd[]) const override;
+
     void lonlat2xy( double crd[] ) const override;
 
     ///< specification for stretching
     Spec spec() const override;
 
-    std::string units() const override { return "meters"; }
+    std::string units() const override { return "degree"; }
```
wdeconinck, line 41:
```suggestion
    std::string units() const override { return "degrees"; }
```

src/atlas/projection/detail/StretchRegular.h
```diff
@@ -46,9 +51,9 @@ class StretchLAM final : public ProjectionImpl {
      }
 
     void checkvalue(const double&, const double&) const;
-    double general_stretch (double&, const bool&, int&, const int&, const int&) const;
+    double general_stretch (double&, const bool&, int, const int, const int) const;
```
wdeconinck, line 54: Passing something by reference makes the argument be of 64bit (for 64bit architectures).
For this reason it was good to pass int (32bit) by value indeed.
You could do the same with the `bool` to be only a 1bit argument.
The first argument is passed as `double&` which indicates it is a "INTENT(INOUT)" variable (Fortran analogy) whereas it should really be a "INTENT(IN)" which you can specify by passing as const-reference `const double&` or by value `const double` as the others. I prefer the latter for consistency with other arguments and because a value (double) or reference are both 64 bit.

```suggestion
    double general_stretch (const double, const bool, int, const int, const int) const;
```
And make the matching change in StretchRegular.cc of course.

wdeconinck (changes requested):

src/tests/grid/test_stretchedRegular.cc
MO-marcomilan: Hi, thank you for the comment.

A comment was missing in StretchRegular.cc, it is a part of the original algorithm (from Terry Davies).
/** compute ratio,
           *  change stretching factor so that high and low grids
           *   retain original sizes
           */
          double var_ints_f = (n_int - n_rim_- ints_high)/2.;
          double logr = log(var_ratio_);
          double log_ratio = (var_ints_f - 0.5) * logr;
          double new_ratio = exp(log_ratio / var_ints);

In this case the actual ratio (new_ratio) is around 1.10722.

MO-marcomilan:
From your comment about the inner and outer grid
> Does that mean that we need to "progress" only 8 steps and then the next increment is `5`?
Yes, so that the last stretched cell has a resolution higher than the rim oder the external grid.

src/tests/grid/test_stretchedRegular.cc
wdeconinck: Thank you very much for that addition, and apologies for my ignorance, but could you perhaps also in words explain what following variables indicate?
- var_ints
- var_ints_f
- n_int
- n_rim
- ints_high
Perhaps a sketch or drawing would help?

src/atlas/projection/detail/StretchRegular.cc
MO-marcomilan: done

src/tests/grid/test_stretchedRegular.cc
MO-marcomilan: image

There are some comments near the variable definitions, here a simple sketch (I hope that you can see it)

src/atlas/projection/detail/StretchRegular.h
MO-marcomilan: done

src/tests/grid/test_stretchedRegular.cc
wdeconinck: Thank you that is very helpful

src/atlas/projection/detail/StretchRegular.h
MO-marcomilan: We used the grid in the UKV (UK variable grid resolution). Is "variable" reasonable for you?

MO-marcomilan:
About the reduction of parameter, I am still thinking about this. 
I have:
n_stretchedx_ = ((deltax_all + epstest - lam_hires_size) / delta_high_) - n_x_rim_;
The number of the stretched grid points depends on the  outer bounding box (deltax_all), on the regular internal grid (lam_hires_size) and on the rim (n_x_rim_)
I will try to reduce the number of parameters, but I am not sure if it is possible.

MO-marcomilan:
Hi @wdeconinck , I think I finished to answer to your questions. Can you double check this please?

wdeconinck:
Thank you @MO-marcomilan for this.

I have refactored the code a bit more for performance.
I can merge this in already as a intermediate mile stone after some continuous integration has run successfully.
I will create a follow-up issue to track things that still need to be addressed.

- review configuration options
- example how to create a new grid where you don't a priori know the bounding boxes or total number of divisions.
- adding test with rotation
- implement `lonlat2xy` function
- implement `hash` function

wdeconinck (approved):

MO-marcomilan:
Thank you for this huge amount of work @wdeconinck :-)
I think that I should creat a new variableresolutionprojection.yml file under:
atlas/doc/example-grids
Do have I to create another PR for this or can I do it in this one?

wdeconinck:
> Thank you for this huge amount of work @wdeconinck :-)
> I think that I should creat a new variableresolutionprojection.yml file under:
> atlas/doc/example-grids
> Do have I to create another PR for this or can I do it in this one?

You can create a new PR once I have merged the current status into develop as an intermediate checkpoint.
We can try to meet offline after New Year to agree to a way forward.

MO-marcomilan:
Locally I had a problem in compilation:
In different files the compiler didn't accept the lines:

#if ATLAS_HAVE_TRANS
#include "atlas/library/config.h"
#if ATLAS_HAVE_ECTRANS
#include "ectrans/transi.h"
#else
#include "transi/trans.h"
#endif
#endif

I changed them using:

#if ATLAS_HAVE_TRANS
#include "atlas/library/config.h"
#include "transi/trans.h"

MO-marcomilan:
I will be in A/L until the third of January. I will be happy to meet you from the 4th January onwards.
Thanks
Marco

MarekWlasak:
> Locally I had a problem in compilation: In different files the compiler didn't accept the lines:
> 
> #if ATLAS_HAVE_TRANS #include "atlas/library/config.h" #if ATLAS_HAVE_ECTRANS #include "ectrans/transi.h" #else #include "transi/trans.h" #endif #endif
> 
> I changed them using:
> 
> #if ATLAS_HAVE_TRANS #include "atlas/library/config.h" #include "transi/trans.h"

The issue here is that we will need to set an appropriate environment variable in the bundle to stop this. I will have a look at this today.

wdeconinck:
Thanks Marco for your contribution. It is now squash-merged in develop with commit 8e90840

MO-marcomilan:
Hi Willem,
I am trying to write my .yml file under:
atlas/doc/example-grids/regional_variable_resolution.yml
I am trying to update the uid using:
python update_uid.py <my_path>/atlas/doc/example-grids/
The files are updated but the uid not, I think I am wrongly given the path, as the python script ask as argument:
the path to the "atlas-grids" executable
Which path have I to give?
Thanks in advance,
Marco

wdeconinck:
Hi Marco, the path to atlas-grids executable is found in

    <build-dir>/bin/atlas-grids

after compilation.

MO-marcomilan:
Hi Willelm,
I used it, this morning.
However my uid are not updated. Do you know why?
Thanks,
Marco

wdeconinck:
The yaml file with the grid description optionally needs to contain a "check" section, see e.g. file `regular_lonlat_1.yml`

    # Regular lonlat
    type : "regular_lonlat"
    nx : 128
    ny : 65

    check:
      size : 8320
      lonlat(first) : [0,90]
      lonlat(last)  : [357.188,-90]
      uid : f109da5924c5998453cdb7d5854ebd1a
      bounding_box(n,w,s,e) : [90,0,-90,360]

To know the values to write in the "check" section, you can use:

    atlas-grids regular_lonlat_1.yml --info

The `update_uid.py` script should then modify the "uid" value in this file, only if you modified how it is is computed.
This is typically not something that should be done by you, but rather by myself when I modify something major that would affect all grids.
In your case I would just modify the "uid" value manually for the one grid you are developing.

MO-marcomilan:
Thanks Willelm,
This is making my life much easier :-)
Marco

MO-marcomilan:
> I have refactored the code a bit more for performance. I can merge this in already as a intermediate mile stone after some continuous integration has run successfully. I will create a follow-up issue to track things that still need to be addressed.
> 
>     * example how to create a new grid where you don't a priori know the bounding boxes or total number of divisions.
I don't understand: the projection is created using the bounding boxes of the original grid and the resolution. The number of the points is computed in:
template <typename Rotation>
VariableResolutionProjectionT<Rotation>::VariableResolutionProjectionT(const eckit::Parametrisation& proj_st):

lines 152-183. Am I missing something?

Issue #83, opened 2021-12-02
Add option to reduce fraction of grid points checked when finite element interpolation fails


twsearle:
## Description

The _finite-element_ interpolation type will attempt to check nearest neighbour cells when a target location is not within the cell suggested by the KDTree search (see [FiniteElement.cc#L263](https://github.com/ecmwf/atlas/blob/91ec3629883b216454d776e0b096d48a4950d4bd/src/atlas/interpolation/method/fe/FiniteElement.cc#L263)).

I think this can be useful, as the KDTree is only based upon the cell centres and may not be perfectly accurate for non-rectangular cells. However, the default behaviour is for this check to search 20% of the grid. This massively impacts performance on large grids where the grid is not global and the target point is outside the grid boundary.

I would like to add an option to set the fraction of the grid to check, with a minimum of 0 such that only the nearest neighbour cells are checked.

## Acceptance Criteria (Definition of Done)

It is possible to set the [maxFractionElemsToTry](https://github.com/ecmwf/atlas/blob/91ec3629883b216454d776e0b096d48a4950d4bd/src/atlas/interpolation/method/fe/FiniteElement.cc#L235) through the configuration of an atlas interpolator, and this behaviour is covered by a test.

PR #84, opened 2021-12-02
feature: Add maxFractionElemsToTry option

twsearle:
## Summary of changes

* Make maxFractionElemsToTry an option read from the configuration of an interpolation method
* Add the option to the print of the interpolator
* Add a test for the interpolator print function

I have tested this functionality by searching the print output. This is perhaps not so good as I don't actually test the behaviour of `maxFractionElemsToTry`, however I noticed in the code coverage that there was no coverage of the print function, so this should slightly improve things!

I might also open another small PR in the future to add a test for `NodeColumns` to `NodeColumns` finite-element interpolation if that is OK. 

This is my first attempt at a PR in atlas, and I am not very clear on the use of `atlas::util::option` vs `atlas::util::config`, so please let me know if this needs to be improved.

## Issues

Fixes issue #83

### DIFF
--- a/src/atlas/interpolation/method/fe/FiniteElement.cc
+++ b/src/atlas/interpolation/method/fe/FiniteElement.cc
@@ -132,20 +132,26 @@ struct Stencil {
 void FiniteElement::print(std::ostream& out) const {
     functionspace::NodeColumns src(source_);
     functionspace::NodeColumns tgt(target_);
-    if (not tgt) {
-        ATLAS_NOTIMPLEMENTED;
+    if ( mpi::rank() == 0 ) {
+        out << "atlas::interpolation::method::FiniteElement" << std::endl;
+        out << "maxFractionElemsToTry_: " << maxFractionElemsToTry_;
+        out << " treat_failure_as_missing_value_: " << treat_failure_as_missing_value_ << std::endl;
     }
+    if ( not tgt ) {
+        return;
+    }
+    out << "NodeColumns to NodeColumns stencil weights: " << std::endl;
     auto gidx_src = array::make_view<gidx_t, 1>(src.nodes().global_index());
 
     ATLAS_ASSERT(tgt.nodes().size() == idx_t(matrix_->rows()));
 
 
     auto field_stencil_points_loc  = tgt.createField<gidx_t>(option::variables(Stencil::max_stencil_size));
-    auto field_stencil_weigths_loc = tgt.createField<double>(option::variables(Stencil::max_stencil_size));
+    auto field_stencil_weights_loc = tgt.createField<double>(option::variables(Stencil::max_stencil_size));
     auto field_stencil_size_loc    = tgt.createField<int>();
 
     auto stencil_points_loc  = array::make_view<gidx_t, 2>(field_stencil_points_loc);
-    auto stencil_weights_loc = array::make_view<double, 2>(field_stencil_weigths_loc);
+    auto stencil_weights_loc = array::make_view<double, 2>(field_stencil_weights_loc);
     auto stencil_size_loc    = array::make_view<idx_t, 1>(field_stencil_size_loc);
     stencil_size_loc.assign(0);
 
@@ -232,7 +238,6 @@ void FiniteElement::setup(const FunctionSpace& source) {
     array::ArrayView<double, 2> out_lonlat = array::make_view<double, 2>(target_lonlat_);
 
     idx_t Nelements                    = meshSource.cells().size();
-    const double maxFractionElemsToTry = 0.2;
 
     // weights -- one per vertex of element, triangles (3) or quads (4)
 
@@ -241,7 +246,7 @@ void FiniteElement::setup(const FunctionSpace& source) {
 
     // search nearest k cell centres
 
-    const idx_t maxNbElemsToTry = std::max<idx_t>(8, idx_t(Nelements * maxFractionElemsToTry));
+    const idx_t maxNbElemsToTry = std::max<idx_t>(8, idx_t(Nelements * maxFractionElemsToTry_));
     idx_t max_neighbours        = 0;
 
     std::vector<size_t> failures;
--- a/src/atlas/interpolation/method/fe/FiniteElement.cc
+++ b/src/atlas/interpolation/method/fe/FiniteElement.cc
@@ -132,11 +132,9 @@ struct Stencil {
 void FiniteElement::print(std::ostream& out) const {
     functionspace::NodeColumns src(source_);
     functionspace::NodeColumns tgt(target_);
-    if ( mpi::rank() == 0 ) {
-        out << "atlas::interpolation::method::FiniteElement" << std::endl;
-        out << "maxFractionElemsToTry_: " << maxFractionElemsToTry_;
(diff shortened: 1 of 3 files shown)


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf/atlas?pullRequest=84) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf/atlas?pullRequest=84) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf/atlas?pullRequest=84) it.</sub>

src/atlas/interpolation/method/fe/FiniteElement.cc
```diff
@@ -132,20 +132,26 @@ struct Stencil {
 void FiniteElement::print( std::ostream& out ) const {
     functionspace::NodeColumns src( source_ );
     functionspace::NodeColumns tgt( target_ );
+    if ( mpi::rank() == 0 ) {
+        out << "atlas::interpolation::method::FiniteElement" << std::endl;
+        out << "maxFractionElemsToTry_: " << maxFractionElemsToTry_;
+        out << " treat_failure_as_missing_value_: " << treat_failure_as_missing_value_ << std::endl; 
+    }
```
wdeconinck, line 139: You can remove "mpi::rank() == 0"
This should be handled at the calling site of this function.

src/atlas/interpolation/method/fe/FiniteElement.h
```diff
@@ -28,7 +28,9 @@ namespace method {
 
 class FiniteElement : public Method {
 public:
-    FiniteElement( const Config& config ) : Method( config ) {}
+    FiniteElement( const Config& config ) : Method( config ) {
+      if (config.has("maxFractionElemsToTry") ) { config.get("maxFractionElemsToTry", maxFractionElemsToTry_); }
```
wdeconinck, line 32: The convention is to use snake_case for the configuration keys, so "max_fraction_elems_to_try".

src/tests/interpolation/test_interpolation_finite_element.cc
```diff
@@ -51,28 +51,38 @@ CASE( "test_interpolation_finite_element" ) {
 
     auto func = []( double x ) -> double { return std::sin( x * M_PI / 180. ); };
 
-    Interpolation interpolation( option::type( "finite-element" ), fs, pointcloud );
+    util::Config config = util::Config( "type", "finite-element" ).set( "maxFractionElemsToTry", 0.4 );
```
wdeconinck, line 54: also here "max_fraction_elems_to_try"

src/atlas/interpolation/method/fe/FiniteElement.cc
```diff
@@ -132,20 +132,26 @@ struct Stencil {
...
     if ( not tgt ) {
-        ATLAS_NOTIMPLEMENTED;
+        return;
```
wdeconinck, line 141: Is this something you encountered?
OK to return, but perhaps output that "tgt" is an incompatible function space or so.

twsearle: yes, the tests actually work by interpolating NodeColumns to PointCloud, which means if I run the print function as part of the test I hit this exception.

 My thinking was the rest of the function only makes sense for printing NodeColumns, however this is not the only compatible target function space.
 
  In fact as far as I can tell there is no test for the NodeColumns to NodeColumns interpolation as it stands. I think it would be good to add one, but I couldn't think of nice simple test points off the top of my head! I will have another think about this if you like!

src/atlas/interpolation/method/fe/FiniteElement.cc
twsearle: Sorry still getting used to this! I have tried to resolve this in the last commit, but it doesn't seem to show you that inline...

src/tests/interpolation/test_interpolation_finite_element.cc
twsearle: should be fixed inthe most recent commit.

src/atlas/interpolation/method/fe/FiniteElement.h
twsearle: should be fixed in the most recent commit.

src/atlas/interpolation/method/fe/FiniteElement.cc
wdeconinck: Thanks for elaborating. I'm happy for you to leave it as is, without any further warning or exception.

src/atlas/interpolation/method/fe/FiniteElement.h
```diff
@@ -28,7 +28,9 @@ namespace method {
 
 class FiniteElement : public Method {
 public:
-    FiniteElement(const Config& config): Method(config) {}
+    FiniteElement(const Config& config): Method(config) {
+        if (config.has("max_fraction_elems_to_try") ) { config.get("max_fraction_elems_to_try", maxFractionElemsToTry_); }
```
wdeconinck, line 32: You can just use
```
config.get("max_fraction_elems_to_try", maxFractionElemsToTry_);
```
The default value is already initialised automatically with line 84, and 'config.get(key,value)' will only change the value if the key is present.

src/tests/interpolation/test_interpolation_finite_element.cc
```diff
@@ -43,28 +43,38 @@ CASE("test_interpolation_finite_element") {
 
     auto func = [](double x) -> double { return std::sin(x * M_PI / 180.); };
 
-    Interpolation interpolation(option::type("finite-element"), fs, pointcloud);
+    util::Config config = util::Config("type", "finite-element").set("max_fraction_elems_to_try", 0.4);
+    Interpolation interpolation(config, fs, pointcloud);
```
wdeconinck, line 47: You can use e.g.
```c++
Interpolation interpolation{option::type("finite-element")|util::Config("max_fraction_elems_to_try",0.4), fs, pointcloud};
```
You can concatenate configurations with the `|` symbol and `option::type("finite-element")` is a shorthand for `util::Config("type","finite-element")`

src/atlas/interpolation/method/fe/FiniteElement.h
twsearle: I see, I missed that. I have made the suggested change in the latest commit.

src/tests/interpolation/test_interpolation_finite_element.cc
twsearle: Sounds fine by me, I have made the suggested change in the latest commit.

wdeconinck (approved):
Everything seems OK now. Thank you Toby.
I will submit to our internal continuous integration and squash-merge into develop.

twsearle:
> Everything seems OK now. Thank you Toby. I will submit to our internal continuous integration and squash-merge into develop.

Sorry, I thought I had rebuilt and run the ctests but I must have made a mistake. Looks like the CI picked up an error that should now be corrected in the latest commit.

wdeconinck:
Thanks for your contribution. It is now merged in develop with commit 86595a4

PR #86, opened 2022-01-12
Cubed sphere tidy up

odlomax:
This is a tidy-up of the cubed-sphere mesh generator and functionspace classes.

Now:
 * Any cubed sphere functionspace (cells or nodes) can be upcast to a `FunctionSpace` class with a lonlat and a ghost field.
 * For a primal cubed sphere mesh with a halo size _n_, a function space can be created with a halo size up to and including _n_. For example, you can create a `CubedSphereCellColumns` functionspace with a halo of 1 from a cubed sphere mesh with a halo of 3. (Note, the cubed sphere dual mesh halo is still fixed).
 * I've also made sure that the cubed sphere functionspaces have a `size` and `sizeOwned` method similar to `StructuredColumns`.
 
 I also made a couple of changes to the CellColumns class:
 * It returns a ghost field if one is present in `mesh().cells()`.
 * It can make a use the mesh metadata to make a zero-halo function space from a mesh with a halo.


wdeconinck:
Merge / rebased in develop

PR #87, opened 2022-01-26
Vortex Rollup

MarekWlasak:
@wdeconinck - This is a small belated Christmas present refactoring the vortex rollup test so that it can be used elsewhere.

### DIFF
--- a/src/atlas/util/VortexRollup.cc
+++ b/src/atlas/util/VortexRollup.cc
@@ -1,0 +1,43 @@
+/*
+ * (C) Crown Copyright 2022 Met Office
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation
+ * nor does it submit to any jurisdiction.
+ */
+
+#include "atlas/util/VortexRollup.h"
+
+namespace atlas {
+
+namespace util {
+
+double vortex_rollup(double lon, double lat, double t,
+                     const double mean, const bool degreesInput,
+                     const double timescale) {
+
+  if (degreesInput) {
+    lon *= Constants::degreesToRadians();
+    lat *= Constants::degreesToRadians();
+  }
+
+  auto sqr           = [](const double x) { return x * x; };
+  auto sech          = [](const double x) { return 1. / std::cosh(x); };
+  const double Omega = 2. * M_PI / timescale;
+  t *= timescale;
+  const double lambda_prime = std::atan2(-std::cos(lon - Omega * t), std::tan(lat));
+  const double rho = 3. * std::sqrt(1. - sqr(std::cos(lat)) * sqr(std::sin(lon - Omega * t)));
+  double omega     = 0.;
+  double a         = Earth::radius();
+  if (rho != 0.) {
+    omega = 0.5 * 3 * std::sqrt(3) * a * Omega * sqr(sech(rho)) * std::tanh(rho) / rho;
+  }
+  double q = mean - std::tanh(0.2 * rho * std::sin(lambda_prime - omega / a * t));
+  return q;
+}
+
+}
+
+}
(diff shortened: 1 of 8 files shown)


wdeconinck:
Hi Marek, good idea to do this. It has been on my wish list for a while.

A few remarks on the new implementation:
- Remove `degreesInput` argument. The contract should just assume degrees. Conversion from radians can be done outside this function if needed.
- Remove `mean` argument: we can always use it as:
  `double field = mean + vortex_rollup(lon,lat,t)`
- Remove `timescale` argument and use within the code as I noticed that it is meaningless. It cancels out in the computations, but please double check!
    - `Omega` is divided with timescale
    - `t` is multiplied with timescale
    - both `Omega` and `t` are always used as `Omega * t` (or via combination with lower-case `omega`)

The final function signature can then be
```
// lon in degrees
// lat in degrees
// values centred around mean = 0
double vortex_rollup(const double lon, const double lat, const double t);
```
I prefer this without optional arguments.

wdeconinck (changes requested):
Please address above comments

src/atlas/util/VortexRollup.h
```diff
@@ -0,0 +1,44 @@
...
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation
+ * nor does it submit to any jurisdiction.
+ */
+
+#pragma once
+
+#include "atlas/util/Constants.h"
+#include "atlas/util/Earth.h"
```
wdeconinck, line 14: Please move these headers to VortexRollup.cc

src/sandbox/interpolation/atlas-parallel-interpolation.cc
```diff
@@ -245,7 +223,7 @@ int AtlasParallelInterpolation::execute(const AtlasTool::Args& args) {
             //            src_scalar_1( j ) = -std::tanh( y / 10. * std::cos( 50. / std::sqrt( x * x + y * y ) ) -
             //                                            x / 10. * std::sin( 50. / std::sqrt( x * x + y * y ) ) );
 
-            src_scalar_1(j) = vortex_rollup(lon, lat, 1.);
+            src_scalar_1(j) = util::vortex_rollup(lon, lat, 1., 1., false);
```
wdeconinck, line 226: Because of the lon, lat in radians, here we better pass `lonlat(j,0)` and `lonlat(j,1)` in degrees

src/atlas/util/VortexRollup.cc
```diff
@@ -0,0 +1,43 @@
...
+                     const double timescale) {
+
+  if (degreesInput) {
+    lon *= Constants::degreesToRadians();
+    lat *= Constants::degreesToRadians();
+  }
+
+  auto sqr           = [](const double x) { return x * x; };
+  auto sech          = [](const double x) { return 1. / std::cosh(x); };
+  const double Omega = 2. * M_PI / timescale;
```
wdeconinck, line 28:
```
constexpr double two_pi = 2. * M_PI;
```
Given that the timescale is meaningless (see other comments), we could just keep using `two_pi` instead of `Omega`

wdeconinck (changes requested):
I realise that the original version was coded by myself, but now that it becomes more official as part of the API we can clean it up.

src/atlas/util/VortexRollup.cc
MarekWlasak: Done

src/sandbox/interpolation/atlas-parallel-interpolation.cc
MarekWlasak: Done

src/atlas/util/VortexRollup.h
MarekWlasak: Done

MarekWlasak:
@wdeconinck - All done!

wdeconinck:
Everything merged in develop with commit [6b508f0](https://github.com/ecmwf/atlas/commit/6b508f02a5d27d15ec8dc93e39a090ed936ebeda)
Thanks Marek!

PR #88, opened 2022-01-28
Feature/ecmwf ukv yaml

MO-marcomilan:
From
https://github.com/ecmwf/atlas/pull/82
As requested from Willelm I added to the stretched grid:
a test with rotation
implemented lonlat2xy function
a test of inverted rotated function


wdeconinck:
Hi @MO-marcomilan 
I have [fixed](https://github.com/ecmwf/atlas/pull/88/commits/bb83a36ce423cec72fdc0b3f212461424343b6e5) the yaml file so that the test passes. It now reads
```yaml
type : "regional"
nx : 34
ny : 32
xmin : 348.13120344827576
xmax : 369.6190965517242
ymin : -8.264495551724226
ymax : 11.92110100000127
projection :
    type: "variable_resolution"
    var_ratio : 1.13
    delta_low : 1.
    delta_hi : 0.6511482758621128
    x_reg_start : 351.386944827586319
    x_reg_end : 366.363355172413776
    y_reg_start : -5.008754172413662
    y_reg_end : 8.665359620690706
    startx : 348.13120344827576
    endx : 369.6190965517242
    starty : -8.264495551724226
    endy : 11.92110100000127
    rim_widthx : 4
    rim_widthy : 4
```
As I mentioned before, this is actually quite overdetermined and in the end I would like something like the following to work "somehow" and we can discuss further how offline.
```yaml
type : "regional_variable_resolution"
progression: 1.13
inner:
    xmin: 351.386944827586319
    ymin: -5.008754172413662
    dx: 0.6511482758621128
    dy: 0.6511482758621128
    nx: 24
    ny: ... # don't know now
outer:
    xmin:  348.13120344827576
    xmax: 369.6190965517242
    ymin: -8.264495551724226
    ymax: 11.92110100000127
    dx: 1.
    dy: 1.
projection : # optional
    type       : "rotated_lonlat"  # Only rotation could be supported at the moment
    north_pole : [-176,40]
```
Foremost take note of the special grid type "regional_variable_resolution". It is in reality still a atlas StructuredGrid, but this type indicates a special grid builder class to help you construct the grid easier.
I believe the total `nx`,`ny`, `inner.xmax`, `inner.ymax`, `outer.rim_width` can somehow be deduced from these keys. Specific code in the grid builder can take care of this.

An alternative way that could coincide as well for instance could be:
```yaml
type : "regional_variable_resolution"
progression: 1.13
inner:
    xmin: 351.386944827586319
    ymin: -5.008754172413662
    dx: 0.6511482758621128
    dy: 0.6511482758621128
    nx: 24
    ny: ... # don't know now
outer:
    nx_rim: 5  # number of points going outwards from inner, on one side in x-direction
    ny_rim: 5  # number of points going outwards from inner, on one side in y-direction
    dx: 1.
    dy: 1.
projection : # optional
    type       : "rotated_lonlat"  # Only rotation could be supported at the moment
    north_pole : [-176,40]
```
In this case you would not specify exactly the outer bounds of the grid, and they will be computed.
- Figure out bounds of transition zone until outer.dx is reached (call it transition.xmax)
- nx_regular_outer = nx_rim - nx_transition   (values one-sided)
- outer.xmax = transition.xmax + outer.dx * nx_regular_outer

All projection parameters are then computed and the projection is created together with the grid

@MarekWlasak @twsearle @odlomax , do you think this is sensible as well?

You could also describe the inner grid differently:
- Deducing inner resolution
  ```yaml
  inner:
      xmin: ...
      xmax: ...
      ymin: ...
      ymax: ...
      nx: ...
      ny: ...
    ```
- Deducing inner number of points
  ```yaml
  inner:
      xmin: ...
      xmax: ...
      ymin: ...
      ymax: ...
      dx: ...
      dy: ...
    ```
- Specify centre, resolution and number of points
  ```yaml
  inner:
      centre(x,y): [ x0, y0 ]
      dx: ...
      dy: ...
      nx: ...
      ny: ...
    ```
- ...

wdeconinck:
The yaml file has now been included with commit [e447b28](https://github.com/ecmwf/atlas/commit/e447b2858474618b2bd29916036ebb71df56d63d)

MO-marcomilan:
Hi Willem,
I started to do this work. I will do stretch as a new grid class instead as a projection.
Is it this OK?
Thanks,
Marco

wdeconinck:
> Hi Willem,
> I started to do this work. I will do stretch as a new grid class instead as a projection.
> Is it this OK?
> Thanks,
> Marco

Hi Marco,
You don't need to create a new grid. Rather you need to create a new grid builder.
See e.g. https://github.com/ecmwf/atlas/blob/develop/src/atlas/grid/detail/grid/Regional.cc
We can discuss offline if you want in the Slack channel.

PR #89, opened 2022-01-28
Feature/ecmwf ukv var

MO-marcomilan:
Add test for rotated grid
Add inverse of projection, from stretched grid to regular grid
Add test of the inverse


MO-marcomilan:
I added a new test in the variable resolution. Using a function which create the points in the variable grid.

Issue #68, opened 2021-08-06
Determine grid resolution specific parts in cubed-sphere grid from attributes and remove xyt space


MarekWlasak:
To be done after we have static B working with trans and the cubed-sphere.

PR #90, opened 2022-02-02
Add extended cubed-sphere dual halo with a correct alpha-beta projection.

odlomax:
This is PRs [86](https://github.com/JCSDA-internal/atlas/pull/86) and [87](https://github.com/JCSDA-internal/atlas/pull/87) from JCSCA-internal, merged into one. The functionality added is as follows:

## [PR 86](https://github.com/JCSDA-internal/atlas/pull/86)

This PR add functionality to the dual mesh halo. Now, the halo size can be set to any value < N-1, where N is the grid-size of a cubed sphere tile. This allows arbitrary-sized stencil operations to work on the cubed-sphere dual mesh.

An example of a cubed sphere dual mesh partition (8 PEs, default partitioner) with a halo of 3 is shown below.

dual-halo-2d
xy projection of an N48 partition with a halo of 3.

dual-halo-3d
xyz projection of an N48 partition with a halo of 3.

I've updated several of the existing cubed sphere dual mesh tests to use larger halo sizes.

## [PR 87](https://github.com/JCSDA-internal/atlas/pull/87)

This PR predominantly adds an `(x, y)` to `(alpha, beta)` projection for tile `t`. Here, `(alpha, beta)` is defined _strictly_ as the angular coordinates described by [Ronchi et al., (1996)](https://doi.org/10.1006/jcph.1996.0047).

In practice, the transform is a simple shift and rotate for `xy` coordinates internal to tile `t`. A non-linear stretch is applied for external points (i.e. halo `xy` coordinates).

The following example considers tile `t=0` of the LFRic cubed sphere dual mesh, with `halo = 3`. The first image shows the `xy` projection. The second image gives the `(alpha, beta)` projection.

dual_halo_before
dual_halo_after

This required changes to several classes, summarised as follows:

 * **grid::Tiles** added `tileCenter` and `tileJacobian` which return the `xy` centre and `d_xy/d_alphabeta` Jacobian for each tile.
 * **meshgenerator::detail::cubedsphere::Jacobian2** removed class and added equivalent to `util` namespace.
 * **meshgenerator::detail::cubedsphere::NeighbourJacobian** updated this class to use the new `Tiles` methods. This minor refactoring therefore serves as a test for the changes.
 *  **CubedSphereGrid** added method to return a reference to the stored `CubedSphereProjectionBase `
 * **redistribution::detail::RedistributionGeneric** removed `rank <= 3` constraint. Not related to projection, but it was an easy fix.

This transform is currently only implemented for the equiangular cubedsphere projection. It is also likely that I've introduced some unnecessary redundancy to the `CubedSphereProjectionBase` and `Tiles` classes. However, given that we intend to [refactor the cubed sphere grid](https://github.com/ecmwf/atlas/issues/68) at some point, we'll eventually need to refactor these classes too.


wdeconinck:
Thanks @odlomax  I will have look very soon

odlomax:
@wdeconinck, after our chat the other day, I removed `JacobianXY` class and added a general matrix class in `atlas/util/Matrix.h`. As per your suggestion, it uses `eckit::maths::Matrix` as a back end.

I also learned (the hard way) that `composition > inheritance`.

odlomax:
Hi @wdeconinck ,

I can't make sense of these linker errors when I try and build without Eigen:

```
/usr/bin/ld: lib/libatlas.so.0.27: undefined reference to `eckit::maths::lapack::getrf(int*, int*, double*, int*, int*, int*)'
  /usr/bin/ld: lib/libatlas.so.0.27: undefined reference to `eckit::maths::lapack::getri(int*, double*, int*, int*, double*, int*, int*)
  ```
  
  Is it anything you've come across before?

odlomax:
> Hi @wdeconinck ,
> 
> I can't make sense of these linker errors when I try and build without Eigen:
> 
> ```
> /usr/bin/ld: lib/libatlas.so.0.27: undefined reference to `eckit::maths::lapack::getrf(int*, int*, double*, int*, int*, int*)'
>   /usr/bin/ld: lib/libatlas.so.0.27: undefined reference to `eckit::maths::lapack::getri(int*, double*, int*, int*, double*, int*, int*)
> ```
> 
> Is it anything you've come across before?

My bad. Fixed it!

PR #91, opened 2022-02-10
Added matrix class.

odlomax:
This adds a Matrix class which uses `eckit::maths::Matrix` as a back end for most of its methods. The class also includes methods for matrix-scalar and matrix-KPoint multiplication, as well as a sign(matrix) method.

The class is designed to provide type-safety when performing arithmetic operations, e.g., the output of a matrix-multiplication is always an `atlas::util::Matrix` object, instead of -- for example -- an Eigen expression.


odlomax:
I've now re-written this into a `KJacobian` class (the name differentiates it from the lonlat Jacobian in `atlas::projection`.). This inherits `eckit::maths::Matrix` and adds a list constructor and a means to perform Jacobian-kPoint multiplications.

wdeconinck:
Some offline discussion conclusion:
Instead, extend Projection::Jacobian to cater for the needs at hand. A new PR will be created for this.

PR #92, opened 2022-02-17
Feature/ecmwf ukv varres yml

MO-marcomilan:
I added Regional_var_resolution.cc. Which contains the variable resolution divided in different parts (inner, outer and optionally the projection for the rotation). The grid is a type of StruturedGrid which uses the VariableResolution projection class.
The parameter definitions is temporary and reflect the yml file. If we will decide to change the yml file the "regional_variable_resolution" grid has to be changed as well. However, the basic structure is coded and a change should be easy to implement.

The projection is described under:
https://github.com/ecmwf/atlas/pull/82

The description of the possible .yml files is under
https://github.com/ecmwf/atlas/pull/88#issuecomment-1030293197

I decided to create a pull request under JCDA-internal because the final form of the .yml file have been not defined.


wdeconinck (approved):
Hi @MO-marcomilan 
This looks great! (Almost) Exactly as we discussed.
I just did some minor modifications in the last commit "Finishing touches"
See if you're happy with it.

MO-marcomilan:
Thanks Willem! :-)

wdeconinck:
Thank you @MO-marcomilan this is now merged in develop with commit a2a53f02c

PR #94, opened 2022-02-18
Remove rank 3 restriction of redistribution class.

odlomax:
This is a minor feature update which removes the field rank restrictions in the redistribution class. Before, the class could only redistribute fields of rank 3, or lower. Now, it can redistribute all ranks supported by Atlas.


wdeconinck (approved):
Looks great to me!

PR #85, opened 2022-01-04
feature: bilinear-remapping interpolation

twsearle:
### Description

This change is to add the NEMOVAR and NEMO bilinear-remapping interpolation scheme to interpolate model data. This scheme is quite similar to the finite-element scheme already available in atlas, however, it operates in 2D using the longitude latitude data. It is not possible to use the "StructuredColumns" methods for this, as the orca grid data is essentially unstructured.

###  Summary of changes
* Introduce 2D triangular and quadrilateral element classes including intersection, area and validity checks.
* Add function to construct cell centres for a 2D grid analogous to BuildCellCentres, for use in the  construction of a 2D KDtree.
* Add PointIndex2 class for the 2D kdtree of the nearest cell centre to each observation in latitude-longitude space.
* add tests for the above

### Issues

Part of issue #4

### Known problems

* This is not quite working as expected with atlas-orca (half the field is missing). I believe this is due to a mismatch between the longitude range in source and target data. Below is an image of the issue for orca data interpolated to the longitude range `[-180,180]`.
bil_orca_sea_ice_hofx
I have a fix for this issue in wdeconinck/atlas-orca#2, however, this needs further investigation as it may not be the best way to go about it. No matter what, I feel this issue is probably best resolved within atlas-orca as it seems to be specific to that mesh.
* I am not sure if this is particularly efficient, however, so far the finite-element method has proven just about efficient enough for our application so this should be OK
* It might be nice to add a good test from NodeColumns -> NodeColumns interpolation - currently I am not sure which grids are different in which areas to make a valid test - however I have one in my atlas-orca PR

PR #96, opened 2022-03-04
fix: Catch periodicity issues in BilinearRemapping interpolation

twsearle:
### Description

PR  #85 exposed an issue with this interpolation method. I thought I had a solution for this in wdeconinck/atlas-orca#2, however, upon further discussion it became clear that this approach is flawed and insufficient to account for the perodicity in the grid.

This fix should remedy problems with the grid periodicity, as well as falling back to inverse distance weighting when all else fails. This matches the behaviour of the similar NEMOVAR interpolation.

### Summary of Changes 

  * Use the 3D kdtree to locate 8 nearest grid elements
  * retry observations which fail in case they are in a cell which overlaps the eastern edge of the grid
  * If all else fails, trust the closest cell is the one found by the 3D KDTree and perform an inverse distance weighted interpolation

### Issues

Part of issue #4

### DIFF
--- a/src/atlas/interpolation/element/Quad2D.h
+++ b/src/atlas/interpolation/element/Quad2D.h
@@ -10,6 +10,7 @@
 #include <iosfwd>
 #include <limits>
 
+#include "atlas/runtime/Exception.h"
 #include "atlas/interpolation/Vector2D.h"
 #include "atlas/interpolation/method/Intersect.h"
 #include "atlas/util/Point.h"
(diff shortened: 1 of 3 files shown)


wdeconinck:
Thanks @twsearle . This indeed is much more robust and would fix the atlas-orca issue.
I have added the contribution to the develop branch with commit [7c0300e](https://github.com/ecmwf/atlas/commit/7c0300e1f3e8b841dabf53f63b0eb4178f6a056f)

PR #97, opened 2022-03-10
Add support for rank 2 fields in the 'missing' nonlinear interpolator

twsearle:
### Description of the issue

Interpolation is failing for the `nonlinear::missing` [interpolators](https://github.com/ecmwf/atlas/blob/develop/src/atlas/interpolation/nonlinear/Missing.h) when the source field has vertical levels. In this case, any view of the field is rank 2, with the second dimension indexing over the level number. The example interpolation fails in lines like [Missing.h#L39](https://github.com/ecmwf/atlas/blob/develop/src/atlas/interpolation/nonlinear/Missing.h#L39). I believe this is because there is an assumption that the field has rank 1. For example, when compiling my original test case I would see the following error:
```
In file included from atlas/src/atlas/array/ArrayView.h:18,
                 from atlas/src/atlas/array.h:25,
                 from atlas/src/tests/interpolation/test_interpolation_non_linear.cc:16:
atlas/src/atlas/array/native/NativeArrayView.h: In instantiation of 'void atlas::array::ArrayView<Value, Rank>::check_bounds(Ints ...) const [with Ints = {int}; Value = double; int Rank = 2]':
atlas/src/atlas/array/native/NativeArrayView.h:167:9:   required from 'atlas::array::ArrayView<Value, Rank>::value_type& atlas::array::ArrayView<Value, Rank>::operator()(Idx ...) [with Idx = {int}; Value = double; int Rank = 2; atlas::array::ArrayView<Value, Rank>::value_type = double]'
atlas/src/tests/interpolation/test_interpolation_non_linear.cc:171:20:   required from here
atlas/src/atlas/array/native/NativeArrayView.h:342:38: error: static assertion failed: Expected number of indices is different from rank of array
  342 |         static_assert(sizeof...(idx) == Rank, "Expected number of indices is different from rank of array");
      |                       ~~~~~~~~~~~~~~~^~~~~~~
```

Atlas fields can be scalar or vector fields as well as surface or volume fields. The data layout is such that:
 * Rank=1 -> scalar horizontal field
 * Rank=2, Variables=1 -> scalar columns
 * Rank=2, Variables>1 -> horizontal multiple variables (vector)
 * Rank=3 -> multiple variables in columns 

It is also important for my application that the soluiton is capable of distinguishing between missing values on different levels. For example, my naive first attempt at a solution in this PR I think will lead to data being marked missing at every level if any particular level is considered missing. In ocean applications, a large percentage of the ocean has missing data at one of the deeper levels but not shallower ones.

### Proposed Solution

Originally, we hoped to solve this problem by re-wrapping the raw field data for each level into a new Field. For example:
```C++
const array::ArraySpec spec(array::ArrayShape{src.shape(0)}, array::ArrayStrides{src.shape(1)});
Value* data = const_cast<Value*>(src.array().data<Value>()) + lev;
atlas::Field src_tmp_slice = atlas::Field("s", data, spec);
```
This rank-1 field would be passed to the `Missing` interpolation post-processor and used to update the weights for each level and perform the interpolation independently for each level.

I think there are a couple of issues with this wrapping approach. Instead I have implemented a temporary field buffer, copying the data in and out of a new array. The reasons for this choice are:
  1. I think re-wrapping the data introduces a stride to any operations on the array which would be bad for vectorisation (my understanding is copying into a temporary field will have similar performance to wrapping, but it would be great to be able to test this).
  2. The methods in [Missing.h](https://github.com/ecmwf/atlas/blob/develop/src/atlas/interpolation/nonlinear/Missing.h) index the data using square brackets `[]` rather than round brackets `()`. I believe the square brackets applied to a view from a re-wrapped field don't index in the same way as the view indexed with round brackets. I think the round brackets give the correct result. I am not exactly sure on why this is the case. I am getting a little lost and confused in the detail on the relationship between a view and an array.
  3. Similarly, even after adjusting for the above, the sparse matrix multiply openMP method ([SparseMatrixMultiply_OpenMP.cc](https://github.com/ecmwf/atlas/blob/242ec84223c69f68821dea0fe28d899e5f80d2dc/src/atlas/linalg/sparse/SparseMatrixMultiply_OpenMP.cc#L35)) doesn't result in the answer I would expect. Inserting print statements shows that indexing the view with same index gives a different answer. I am again a bit lost with this, but I think it is to do with the casting of a FieldView to a LocalView in this function.

 * Test cases with rank-2 to rank-2 nonlinear::missing interpolation included pass.
 * missing value interpolation of data with different missing values on different levels.

### Impact

This is an enhancement to atlas, and should be backwards compatible.

### DIFF
--- a/src/atlas/interpolation/method/Method.cc
+++ b/src/atlas/interpolation/method/Method.cc
@@ -132,23 +132,25 @@ void Method::interpolate_field_rank2(const Field& src, Field& tgt, const Matrix&
             // non-linearities: a non-empty M matrix contains the corrections applied to matrix_
             Matrix M;
 
-            const array::ArraySpec spec(array::ArrayShape{src.shape(0)}, array::ArrayStrides{src.shape(1)});
-            Value* data = const_cast<Value*>(src.array().data<Value>()) + lev;
-            atlas::Field src_slice = atlas::Field("s", data, spec);
-            src_slice.metadata() = src.metadata();
-            if (nonLinear_->execute(W_nl, src_slice)) {
+            const array::DataType dtype = array::DataType::create<Value>();
+            atlas::Field src_tmp_slice = atlas::Field("s", dtype, array::ArraySpec(array::ArrayShape{src.shape(0)}));
+            if (src.metadata().has("missing_value"))
+               src_tmp_slice.metadata().set("missing_value", src.metadata().get<Value>("missing_value"));
+            if (src.metadata().has("missing_value_type"))
+               src_tmp_slice.metadata().set("missing_value_type", src.metadata().getString("missing_value_type"));
+            auto src_tmp_slice_v = array::make_view<Value, 1>(src_tmp_slice);
+            for (idx_t i = 0; i < src.shape(0); ++i) {
+                src_tmp_slice_v(i) = src_v(i, lev);
+            }
+            if (nonLinear_->execute(W_nl, src_tmp_slice)) {
                 M.swap(W_nl);
             }
-            // would be nice to do it the following way, but this is incompatible with sparse_matrix_multiply
-            //auto src_v_slice = src_v.slice(array::Range::all(), lev);
-            //auto tgt_v_slice = tgt_v.slice(array::Range::all(), lev);
-            const array::ArraySpec spec_t(array::ArrayShape{tgt.shape(0)}, array::ArrayStrides{tgt.shape(1)});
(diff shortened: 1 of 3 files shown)


wdeconinck:
Hi @twsearle I have reviewed, adapted and squash-merged your contribution into develop with commit https://github.com/ecmwf/atlas/commit/a1916fd521f9935f846004e6194f80275de4de83
Thank you for adding a unit test for this as well!

PR #98, opened 2022-03-16
Cubed sphere bilinear interpolation

odlomax:
This feature adds following functionality:

* a "cubedsphere-bilinear" interpolation scheme, with an associated "cubedsphere" MatchingMesh partitioner.
* adjoint-of-halo-exchange functionality to the NodeColumns functionspace.
* a general tidy-up of Quad2D and Triag2D interpolation elements, with some extra floating-point epsilon checks.
* a cubed sphere interpolation test which uses all of the above. (Example shown below).

CS-LFR-24 interpolation source field:
source

O24 interpolation target field:
target

@wdeconinck and @MarekWlasak I basically copy-and-pasted code from StructuredColumns to NodeColumns to get the adjoint of the halo exchange to work. Can you check that I haven't done something truly daft?


src/tests/interpolation/CMakeLists.txt
```diff
@@ -97,3 +97,11 @@ ecbuild_add_test( TARGET atlas_test_interpolation_structured2D_to_unstructured
   CONDITION eckit_HAVE_MPI
   ENVIRONMENT ${ATLAS_TEST_ENVIRONMENT}
 )
+
+ecbuild_add_test( TARGET atlas_test_interpolation_cubedsphere
+  SOURCES  test_interpolation_cubedsphere.cc
+  LIBS     atlas
+  MPI      6
+  CONDITION eckit_HAVE_MPI AND MPI_SLOTS GREATER_EQUAL 6 AND eckit_EIGEN_FOUND
```
wdeconinck, line 105: Why is eckit_EIGEN_FOUND important?

wdeconinck (changes requested):
Thank you @odlomax this is very high quality code with only very few suggestions.

I have however concerns about the changes to interpolation/element files:
- changes to solveQuadratic function
- changes to the inTriangle and inQuadrilateral methods
- changes or introduction of epsilons
I am not sure what the repercussions are on computations of intersections.
Could you elaborate why the changes were necessary?
If important, perhaps these changes merit a separate PR for which we need to test carefully the impact on operational product generation results and introduce a backward compatible mode, or be able to cleanly revert.
Requesting @pmaciel input on this as well.

src/tests/interpolation/CMakeLists.txt
odlomax: Whoops. It isn't!

odlomax:
> Thank you @odlomax this is very high quality code with only very few suggestions.
> 
> I have however concerns about the changes to interpolation/element files:
> 
> * changes to solveQuadratic function
> * changes to the inTriangle and inQuadrilateral methods
> * changes or introduction of epsilons
>   I am not sure what the repercussions are on computations of intersections.
>   Could you elaborate why the changes were necessary?
>   If important, perhaps these changes merit a separate PR for which we need to test carefully the impact on operational product generation results and introduce a backward compatible mode, or be able to cleanly revert.
>   Requesting @pmaciel input on this as well.

Thanks, @wdeconinck!

I've rolled back the elements to what they were. You're right, I shouldn't have changed them here. I'll investigate properly and raise an issue if I find something unusual.

wdeconinck (approved):
Thanks for addressing all suggestions!

PR #102, opened 2022-04-06
Added fused multiply-add to discriminant calculation.

odlomax:
@pmaciel @wdeconinck @twsearle 

This is the fused multiply-add operation that corrects for the potential loss of [half the significant figures in the quadratic solve](https://en.wikipedia.org/wiki/Loss_of_significance#A_better_algorithm).

The solution here was basically taken from [this Stack Overflow post](https://stackoverflow.com/questions/63665010/accurate-floating-point-computation-of-the-sum-and-difference-of-two-products)

The problem is that this could potentially be slow if the hardware doesn't support fused multiply-add operations (there doesn't seem to be a consistent macro to check for this either).

Anyway, if you're interested, I'll write a proper test for it. If not, never mind! 😊

p.s., I've just realised this also applies to the determinants of 2x2 matrices! 😮

### DIFF
--- a/src/atlas/interpolation/element/Quad2D.cc
+++ b/src/atlas/interpolation/element/Quad2D.cc
@@ -71,16 +71,25 @@ method::Intersect Quad2D::localRemap(const PointXY& p, double edgeEpsilon, doubl
             //   x1 = (-b - sign(b) sqrt(b^2 - 4ac)) / 2a
             //   x2 = c / ax1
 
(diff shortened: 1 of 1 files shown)


wdeconinck:
Thanks @odlomax this is now merged in develop with commit https://github.com/ecmwf/atlas/commit/a3980f7b2ef6d8a0201ec27a8ba0fc15776c9442
Very ingenious method.

PR #104, opened 2022-05-17
Bugfix/cubed sphere interpolation

odlomax:
Found a couple of nasty bugs that slipped through the tests. 

First one is where the CubedSphereBilinear class builds the interpolation matrix for every single target point!

Second one involves the CubedSphereProjectionBase returning a tiles object by value instead of const reference. @MarekWlasak can you confirm that this behaviour is not intentional?


wdeconinck (approved):
Should give big performance improvement!

odlomax:
> Should give big performance improvement!

It not bug-fixing, it's optimisation! 🤣

Took a while to track down the `getCubedSphereTiles` bug. The `ObjectHandle` was causing a race condition when using OpenMP.

wdeconinck:
Thanks for your changes, this is now merged develop with commit 89229b4

PR #110, opened 2022-09-29
bugfix remove spurious assert 

MarekWlasak:


MarekWlasak:
@wdeconinck - hi there.  For some reason "brew" is not present.  Any ideas?

wdeconinck:
Hi @MarekWlasak not sure what's going on. They must have updated the runners. I need to investigate. This change is harmless though.

PR #114, opened 2022-12-13
Feature: Serial distribution to use entire mesh on every partition

twsearle:
## Description

JEDI-OOPS has a "round-robin" configuration, that evenly and randomly distributes observations across each MPI partition. I would like an atlas distribution to support this configuration for use in embarrassingly parallel applications - where each observation requires no communication with any other. The advantage of this is that it will increase the speed of the program whilst allowing for easy testing against domain-decomposed partitioning.

This change will create a copy of the mesh on each MPI parititon allowing atlas to be run in parallel for this kind of problem. I will then create a PointCloud functionspace containing the observation locations on that particular partition and interpolate the gridded data to these locations.

Please let me know if you think this is a bad idea, or if you would prefer I introduce a new distribution for this task ("parallel distribution?") rather than change the serial distribution.

## Impact

Fundamentally changes the "serial" distribution which might have impacts for current users due to increased memory requirements.

### DIFF
--- a/src/atlas/grid/detail/distribution/SerialDistribution.cc
+++ b/src/atlas/grid/detail/distribution/SerialDistribution.cc
@@ -14,6 +14,7 @@
 #include <ostream>
 
 #include "atlas/grid/Grid.h"
+#include "atlas/parallel/mpi/mpi.h"
 
 namespace atlas {
 namespace grid {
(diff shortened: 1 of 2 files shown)


wdeconinck (commented):
This is perfectly fine. I don't have any problem with each partition having the full mesh.
The original intention was to have a quick partitioner in case mpi_size == 1.

One change I would like to see though:
Since this is a performance critical inline function, I would like to have mpi::rank() be stored as a member variable "int mpi_rank_" instead.

PR #124, opened 2023-04-20
Field clone

ytremolet:
First version of the field clone. The c++ test is passing but there is a failure in a fortran test which doesn't make sense given what has changed. Creating a draft PR in case it's obvious to someone.


ytremolet:
Changed to PR now that it seems to be working. Provides a deep copy in addition to the default shallow copy.

Issue #129, opened 2023-05-02
Development of efficient conservative spherical polygon interpolation


wdeconinck:
###   ✅ Phase 1:  #130 

Problem:
- Polygon intersection not precise for pole points with more than 10k edges going into it, e.g. S43200x21600 source to O1280 target has significant under-coverage of target cells at the poles.

Tasks:

-  Easier diagnosis: create automatic detection of one worst example of a target cell covering to be easily visualised in python polygon viewer.

-  Fix problem by revision of polygon intersection algorithm

### Phase 2: Not started 

Problem:
- The algorithm to find suitable source polygons for intersection makes use of a "compare_pointxyz" function which contains an epsilon that needs to be tweaked depending on grid resolution.
- The same algorithm is expensive and contains an OpenMP "critical" region.

Tasks:
- Make use of deterministic information to detect source polygons. Only source polygons which are not in the periodic halo should be considered.

### Phase 3: Not started

Problem:
- The algorithm loops over source polygons and finds suitable target polygons for intersection. Interpolation weight contributions are then assembled for the target polygons. This prevents efficient OpenMP parallelisation, by introducing an OpenMP "critical" region.
- The resulting vector of `eckit::linalg::Triplet` then contains multiple triplets corresponding to the same row, col of the interpolation matrix. Before creating the interpolation matrix these are now merged into unique contributions, which is very inefficient.

Task:
- Invert the algorithm to loop over target polygons to detect source polygons. Probably a can of worms...

wdeconinck:
Thanks for this work Slavko!

PR #130, opened 2023-05-02
Improve robustness of conservative spherical polygon interpolation for high resolution meshes

sbrdar:
Fixing #129
- improvement in polygon intersection (more robust Sutherland-Hodgman algorithm)
- print out the worst under-covering and over-covering of target polygons
- output in JSON format
Remaining issue from #129
- searching of candidate source polygons to intersect a give target polygon should not be done via std::set and comparison of centroids to determine if two polygons are same. The coefficient ATLAS_COMPAREPOINTXYZ_EPS_FACTOR should be set to 1e4 for high grid res, but 1e8 for low mesh. This has to be removed in the future version.

### DIFF
--- a/src/atlas/interpolation/method/unstructured/ConservativeSphericalPolygonInterpolation.cc
+++ b/src/atlas/interpolation/method/unstructured/ConservativeSphericalPolygonInterpolation.cc
@@ -8,8 +8,10 @@
  * nor does it submit to any jurisdiction.
  */
 
+#include <fstream>
 #include <iomanip>
 #include <vector>
+#include <sys/stat.h> // for mkdir
 
 #include "ConservativeSphericalPolygonInterpolation.h"
 
@@ -22,6 +24,7 @@
 #include "atlas/mesh/actions/BuildNode2CellConnectivity.h"
 #include "atlas/meshgenerator.h"
 #include "atlas/parallel/mpi/mpi.h"
+#include "atlas/parallel/omp/omp.h"
 #include "atlas/runtime/Exception.h"
 #include "atlas/runtime/Log.h"
 #include "atlas/runtime/Trace.h"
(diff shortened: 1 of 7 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/atlas?pullRequest=130) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/atlas?pullRequest=130) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/atlas?pullRequest=130) it.</sub>

sbrdar:
Hi Andrea. Thanks for your good remarks! We do aim for a generic conservative interpolation, MPI-parallel & 1st/2nd order. Data on mesh can be either in the cell centres or on the vertices of a mesh. We intersect only convex polygons and polygon edges being great arcs, so we have to split non-convex polygon around a vertex into convex quads. This has worked well for us so far. I would be grateful if you can provide us with the one example causing problems in the "pre-Sutherland-Hodgman" SCRIP library. Thanks again!

andreapiacentini:
> if you can provide us with the one example

Done!  
At least if you mail has the usual format name.family_name@ecmwf.int  
Let me know if you get the message and the file transfer. The grids coordinates are in a scrip like grids file. There were loader around at the time when I tested Atlas interpolations, but at the time we did not talk much of non-convex dual finite volumes grids!  
Don't hesitate to get in touch for any questions or brainstorming. I can also put you in touch with the FESOM developers: theory certainly know much more than I do about this grids. Sometimes my Italian flavor of written English collides with their German flavor, but it would be easier for you !

wdeconinck (approved):
Great work @sbrdar !!!

PR #133, opened 2023-05-09
Optimise array::helpers::ForEach for small arrays.

### LINKED ISSUE #134: ForEach method very slow on small arrays
The current ForEach method is particularly slow when processing a large number number of small arrays. This is probably due to the overhead of using a config object to to set the execution policy.

Issue addressed by [PR 113](https://github.com/ecmwf/atlas/pull/133).

odlomax:
Hi @wdeconinck,

I've found that the array ForEach method is particularly slow when processing small arrays (tested on N = 100). I believe this is because processing `execution_policy` configs has a relatively high amount overhead.

I've done a couple of things to alleviate this:
1) allowed `constexpr` policy selection when you provide an `execution_policy` object.
2) implemented perfect forwarding on the ArrayView objects (I don't think this did much, but it was useful to learn how to do it!)

I've added two performance tests :
1) `nested` performs a `ForEach` over the 50000 columns and another ForEach over the 100 levels. Execution policy selection is `constexpr`.
2) `nested, config`. Does the same as above, but determines the policy via a config.

```
Elapsed time: Addition; raw pointer               = 0.0091302s
Elapsed time: Addition; for loop (i, j)           = 0.00921452s	;   relative to baseline : 100.924%
Elapsed time: Addition; for loop (j, i)           = 0.0478493s	;   relative to baseline : 524.077%
Elapsed time: Addition; for each (columns)        = 0.00907069s	;   relative to baseline : 99.3483%
Elapsed time: Addition; for each (levels)         = 0.0481194s	;   relative to baseline : 527.036%
Elapsed time: Addition; for each (all elements)   = 0.00914668s	;   relative to baseline : 100.181%
Elapsed time: Addition; for each (nested)         = 0.00920637s	;   relative to baseline : 100.834%
Elapsed time: Addition; for each (nested, config) = 0.118888s	;   relative to baseline : 1302.14%

Elapsed time: Trig    ; raw pointer               = 0.312925s
Elapsed time: Trig    ; for loop (i, j)           = 0.319474s	;   relative to baseline : 102.093%
Elapsed time: Trig    ; for loop (j, i)           = 0.502402s	;   relative to baseline : 160.55%
Elapsed time: Trig    ; for each (columns)        = 0.316493s	;   relative to baseline : 101.14%
Elapsed time: Trig    ; for each (levels)         = 0.435909s	;   relative to baseline : 139.301%
Elapsed time: Trig    ; for each (all elements)   = 0.315607s	;   relative to baseline : 100.857%
Elapsed time: Trig    ; for each (nested)         = 0.318103s	;   relative to baseline : 101.655%
Elapsed time: Trig    ; for each (nested, config) = 0.431594s	;   relative to baseline : 137.923%
```

As you can see, the benefit of using a `constexpr` execution policy is quite dramatic when performing a large number calls on small array sizes.

Thankfully, the config overhead is negligible when dealing with normal-sized NWP fields.

The only (intended) change to the interface is that you can now supply ArrayViews to the ForEach::Apply method using `std::tie` and `std::forward_as_tuple` as well as `std::make_tuple`.

closes #134

### DIFF
--- a/src/atlas/array/helpers/ArrayForEach.h
+++ b/src/atlas/array/helpers/ArrayForEach.h
@@ -7,12 +7,12 @@
 
 #pragma once
 
-#include <functional>
 #include <tuple>
 #include <type_traits>
 #include <string_view>
 
 #include "atlas/array/ArrayView.h"
+#include "atlas/array/IndexView.h"
 #include "atlas/array/Range.h"
 #include "atlas/array/helpers/ArraySlicer.h"
 #include "atlas/parallel/omp/omp.h"
@@ -91,15 +91,6 @@ namespace helpers {
 
 namespace detail {
 
-template <typename...>
-struct is_tuple_impl : std::false_type {};
-
-template <typename... Args>
-struct is_tuple_impl<std::tuple<Args...>> : std::true_type {};
-
-template <typename Tuple>
-using is_tuple = std::enable_if_t<is_tuple_impl<std::decay<Tuple>>::value>*;
-
 template <typename... Ts, typename T>
 constexpr auto tuplePushBack(const std::tuple<Ts...>& tuple, T value) {
   return std::tuple_cat(tuple, std::make_tuple(value));
@@ -132,28 +123,56 @@ constexpr auto argPadding() {
   }
 }
 
-template <typename... SlicerArgs, template <typename, int> typename View,
-          typename Value, int Rank, typename... ArrayViews>
+//template <template <typename, int> typename View, typename Value, int Rank>
+//constexpr auto getRank(const View<Value, Rank>&) {return Rank;}
+
+//template <template <typename, int> typename View, typename Value, int Rank>
+//constexpr auto getRank(const View<Value, Rank>&&) {return Rank;}
+
+template<int Rank>
+struct GetRankImpl {
+  constexpr static int rank = Rank;
+};
+
+template<typename>
+struct GetRank;
+
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<View<Value, Rank>> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<View<Value, Rank>&> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<const View<Value, Rank>&> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<View<Value, Rank>&&> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<const View<Value, Rank>&&> : GetRankImpl<Rank> {};
+
+template <size_t ViewIdx = 0, typename... SlicerArgs, typename ArrayViewTuple>
 auto makeSlices(const std::tuple<SlicerArgs...>& slicerArgs,
-                View<Value, Rank>& arrayView, ArrayViews&... arrayViews) {
+                ArrayViewTuple&& arrayViews) {
 
-  // "Lambdafy" slicer apply method to work with std::apply.
-  const auto slicer = [&arrayView](const auto&... args) {
-    return std::make_tuple(arrayView.slice(args...));
-  };
+  using ViewTupleType = std::decay_t<ArrayViewTuple>;
 
-  // Fill out the remaining slicerArgs with Range::all().
-  constexpr auto Dim = sizeof...(SlicerArgs);
-  const auto paddedArgs = std::tuple_cat(slicerArgs, argPadding<Rank - Dim>());
+  if constexpr(ViewIdx < std::tuple_size_v<ViewTupleType>) {
 
-  if constexpr(sizeof...(ArrayViews) > 0) {
+    auto&& view = std::get<ViewIdx>(arrayViews);
+    using View = std::tuple_element_t<ViewIdx, ViewTupleType>;
 
-    // Recurse until all views are sliced.
+    constexpr auto Dim = sizeof...(SlicerArgs);
+    constexpr auto Rank = GetRank<View>::rank;
+    const auto paddedArgs =
+      std::tuple_cat(slicerArgs, argPadding<Rank - Dim>());
+
+    const auto slicer = [&view](const auto&... args) {
+      return std::make_tuple(view.slice(args...));
+    };
+
     return std::tuple_cat(std::apply(slicer, paddedArgs),
-                          makeSlices(slicerArgs, arrayViews...));
+                          makeSlices<ViewIdx + 1>(slicerArgs, arrayViews));
   }
   else {
-    return std::apply(slicer, paddedArgs);
+    return std::make_tuple();
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -114,28 +129,50 @@ constexpr auto argPadding() {
...
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<View<Value, Rank>> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<View<Value, Rank>&> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<const View<Value, Rank>&> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<View<Value, Rank>&&> : GetRankImpl<Rank> {};
+template<template <typename, int> typename View, typename Value, int Rank>
+struct GetRank<const View<Value, Rank>&&> : GetRankImpl<Rank> {};
```
wdeconinck: I don't think the `const View` overloads are required. `View` itself is possible to be const

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -82,6 +88,15 @@ namespace helpers {
...
 
+template <typename>
+struct IsTupleImpl : std::false_type {};
+
+template <typename... Args>
+struct IsTupleImpl<std::tuple<Args...>> : std::true_type {};
+
+template <typename Tuple>
+using IsTuple = std::enable_if_t<IsTupleImpl<Tuple>::value>*;
+
```
wdeconinck, line 99: You can create a 
```c++
template <typename Tuple>
constexpr bool is_tuple() {
   return IsTupleImpl<Tuple>::value;
}
```
or
```c++
template <typename Tuple>
inline constexpr bool is_tuple_v = IsTupleImpl<Tuple>::value;
```
And then us it as
```c++
template <typename ArrayViewTuple, typename = std::enable_if_t<is_tuple<ArrayViewTuple>()>>
void myfunction(ArrayViewTuple&&, ......) 
```
or
```c++
template <typename ArrayViewTuple, typename = std::enable_if_t<is_tuple_v<ArrayViewTuple>>>
void myfunction(ArrayViewTuple&&, ......) 
```

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -114,28 +129,50 @@ constexpr auto argPadding() {
...
-  constexpr auto Dim = sizeof...(SlicerArgs);
-  const auto paddedArgs = std::tuple_cat(slicerArgs, argPadding<Rank - Dim>());
+  if constexpr(ViewIdx < std::tuple_size_v<ViewTupleType>) {
 
-  if constexpr(sizeof...(ArrayViews) > 0) {
+    auto&& view = std::get<ViewIdx>(arrayViews);
+    using View = std::tuple_element_t<ViewIdx, ViewTupleType>;
+
+    constexpr auto Dim = sizeof...(SlicerArgs);
+    constexpr auto Rank = GetRank<View>::rank;
```
wdeconinck, line 152: Wouldn't this work: 
```c++
constexpr auto Rank = View::rank();
```
or maybe even:
```c++
constexpr auto Rank = view.rank();
```
In that latter case you don't need to define `View` type

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -114,28 +129,50 @@ constexpr auto argPadding() {
...
+  using ViewTupleType = std::decay_t<ArrayViewTuple>;
 
-  // Fill out the remaining slicerArgs with Range::all().
-  constexpr auto Dim = sizeof...(SlicerArgs);
-  const auto paddedArgs = std::tuple_cat(slicerArgs, argPadding<Rank - Dim>());
+  if constexpr(ViewIdx < std::tuple_size_v<ViewTupleType>) {
 
-  if constexpr(sizeof...(ArrayViews) > 0) {
+    auto&& view = std::get<ViewIdx>(arrayViews);
+    using View = std::tuple_element_t<ViewIdx, ViewTupleType>;
```
wdeconinck: I think you could also use this:
```c++
using View = std::decay_t< decltype(view) >;
```
and avoid the need to define `ViewTupleType`, and just use `ArrayViewTuple` instead.

wdeconinck (changes requested):
Nice improvements! I like that you don't have the `arrayViewsCopy` anymore.

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -236,69 +258,83 @@ struct ArrayForEach {
   ///         sequential (row-major) order.
   ///         Note: The lowest ArrayView.rank() must be greater than or equal
   ///         to the highest dim in ItrDims. TODO: static checking for this.
-  template <typename... ArrayViews, typename Mask, typename Function>
+  template <typename ArrayViewTuple, typename Mask, typename Function,
+            detail::IsTuple<ArrayViewTuple> = nullptr>
```
fmahebert, line 262: [not a change request, just looking to learn if possible] If I correctly understand that the intention of this line is to sanity-check that `ArrayViewTuple` is indeed a tuple (I don't see an alternative implementation that is turned on when it's not a tuple, but maybe I just missed it?) ... then would it also work to delete `detail::IsTuple` and replace this last template parameter with a `static_assert(IsTupleImpl<ArrayViewTuple>::value);` in the function body? To me that's easier to understand and has fewer lines of code, but is there a reason to chose one over the other besides aesthetics?

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -59,6 +57,14 @@ constexpr std::string_view policy_name(execution_policy) {
...
 }
 
+// Type check for execution policy (Not in C++ standard)
+template <typename execution_policy>
+using is_execution_policy = std::enable_if_t<
+    std::is_same_v<execution_policy, sequenced_policy> ||
+    std::is_same_v<execution_policy, parallel_policy> ||
+    std::is_same_v<execution_policy, parallel_unsequenced_policy> ||
+    std::is_same_v<execution_policy, unsequenced_policy>>*;
+
```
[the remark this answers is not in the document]
odlomax: Done. That's a much more intuitive way of doing it.

src/atlas/array/helpers/ArrayForEach.h
odlomax: Done. Again, much nicer.

src/atlas/array/helpers/ArrayForEach.h
odlomax: Somehow convinced myself this doesn't work. Turns out I was wrong!

src/atlas/array/helpers/ArrayForEach.h
odlomax: The first one works a treat.

odlomax:
> Nice improvements! I like that you don't have the `arrayViewsCopy` anymore.

Thanks, Willem. Your suggestions make the code a lot cleaner (and hopefully readable). I think I know how forwarding works now!

wdeconinck:
FYI, before you go deeper, I have already finished working on a field::for_each API.
I was just about to create a PR, but then you dropped this. I'll have to reintegrate.

odlomax:
> to create

Whoops, I hope I didn't create much more work. I think I've gone more than deep enough!

wdeconinck (approved):
Fantastic

PR #135, opened 2023-05-09
Add function to build mesh from imported connectivity data

fmahebert:
For use in applications like JCSDA's JEDI where a mesh connectivity has been constructed externally to atlas (e.g. by a forecast model that does not use atlas to construct its mesh), and we want to reuse this connectivity to build an atlas NodeColumns FunctionSpace.

This function takes local (per MPI task) data about the nodes and cell-to-node connectivities, and sets up the corresponding fields in a Mesh.


wdeconinck:
Hi @fmahebert thanks that looks like a good approach.

As a first comment I'm not so keen on `std::vector<QuadConnectivityData>` or `std::vector<TriConnectivityData>`.
Perhaps we could have instead extra arguments like:
`const std::vector<std::array<idx_t,4>>& quad_nodes, const std::vector<gidx_t>& quad_global_index, const std::vector<idx_t>& quad_remote_index`
and similar for `triangle_nodes, triangle_global_index, triangle_remote_index`.

Also, how come you provide  remote_index for cells but not for nodes? Nodes' `remote_index` can also be computed after. Is that what you rely on?  In case of `halo==0` there should also not be the need for cells_remote_index as it matches the local numbering.

fmahebert:
> As a first comment I'm not so keen on std::vector<QuadConnectivityData> or std::vector<TriConnectivityData>.

Sounds good, I will split the structs and pass extra arguments as you suggest.

> Also, how come you provide remote_index for cells but not for nodes?

I don't remember how I got here, most likely ignorance?

What do you think of this plan:
1. Add a new argument passing in the remote_index for the nodes.
2. Remove the remote_index for the cells, because as you say, this code is assuming `halo==0` anyway right now. (In the future the interface could be expanded to allow the user to pass more nodes and more cell connectivities, some of which could be marked as halo...)

Does this match your suggestion or have I misunderstood?

Finally, an unrelated question: what is the atlas policy on copyright headers? should I put UCAR's header?

wdeconinck:
That matches the suggestions indeed.
Yes please add the copyright header of UCAR in new files or files that have a significant contribution.

fmahebert:
Thanks for the feedback @wdeconinck! I've made some updates here, and will check how the new version integrates with fv3-jedi tomorrow. I'll report back then.

fmahebert:
I've tested importing the fv3-jedi grid into an atlas Mesh via the updated interface, and it works just as well (and is easier). Happy on my end.

src/atlas/mesh/BuildMeshFromConnectivities.h
```diff
@@ -0,0 +1,52 @@
...
+ * The tri/quad connectivities (boundary_nodes and global_indices) are vectors ranging over the
+ * cells owned by the MPI task. Each cell is defined by a list of nodes defining its boundary;
+ * note that each boundary node must be locally known (whether as an owned of ghost node on the
+ * MPI task), in other words, must be an element of the node global_indices. The cell global index
+ * is, here also, a uniform labeling over the of the cells across all MPI tasks.
+ *
+ * Some limitations of the current implementation (but not inherent):
+ * - can only set up triangle and quad cells.
+ * - cannot import halos, i.e., cells owned by other MPI tasks; halos can still be subsequently
+ *   computed by calling the BuildMesh action.
```
fmahebert, line 39: @ytremolet the current design of this PR means we'll ask the JEDI models to call the `build_halo` action after creating the mesh but before constructing the functionspace. It's one very simple line of code, so not a big deal. 

We could streamline a little either by...
1. adding a bool option here to call `build_halo` internally to this function
2. expanding the interface so models can import their halos directly
 
Personally I like the current design, because it keeps the responsibilities clear and makes the halo-building step easy to spot. Do you have a different opinion as to whether either of these options is significantly better than the current state?

ytremolet: It seems fine. The test doesn't cover that part (or I missed it). That's ok because it's not part of the new function, but do you have an example (in fv3 or other?) to get a better idea what it looks like from the user point of view?

fmahebert: @ytremolet I don't have something very clean, but on fv3-jedi branch `feature/fmahebert_test_atlas_mesh_builder` you can see (note to others, this is not a public repo) a code block in Geometry.cc that sets up the mesh, builds a halo, and tests some halo exchange; in the fortran you can see how the connectivity is inferred from the grid structure; all the above is mixed in which other small changes I was testing but didn't take the time to clean out (sorry).

ytremolet: I had a look at your example. It looks reasonable given the information we want to get from the models. And for now the halo is not shocking, this is code that is only needed once for each model, it won't be repeated all over the place so I would be happy with the current implementation.

wdeconinck:
Perfect.

Note that we also need to add a "base" argument for remote_indices.
It is "1" if the field is referring to Fortran numbering.
Subtract the base from each value before assigning it to the `ridx` index view.


Now fine-tuning the API.

How about creating a class with operator() in the  atlas::mesh namespace.

Also let's make the implementation using C-style arrays, and have the previous argument list using std::vector's delegate to it. This way we can possibly support different container types, and create Fortran interfaces using Fortran arrays without incurring extra copies.

I mean something like following.
```
class MeshBuilder {
public:
    MeshBuilder(const eckit::Configuration& = util::NoConfig()) {
        // Cater for extra configuration if needed, e.g. MPI communicator in coming development
    }
    
    Mesh operator()(size_t nb_nodes,
                    const double lons[], const double lats[],
                    const int ghosts[], const gidx_t global_indices[],
                    const idx_t remote_indices[], const idx_t base, const int partitions[],
                    size_t nb_tris,  const gidx_t tri_boundary_nodes[],  const gidx_t tri_global_indices[],
                    size_t nb_quads, const gidx_t quad_boundary_nodes[], const gidx_t quad_global_indices[]) {
        // THE IMPLEMENTATION
    }

    Mesh operator()(const std::vector<double>& lons, const std::vector<double>& lats,
                    const std::vector<int>& ghosts, const std::vector<gidx_t>& global_indices,
                    const std::vector<idx_t>& remote_indices, const idx_t base, const std::vector<int>& partitions,
                    const std::vector<std::array<gidx_t, 3>>& tri_boundary_nodes,
                    const std::vector<gidx_t>& tri_global_indices,
                    const std::vector<std::array<gidx_t, 4>>& quad_boundary_nodes,
                    const std::vector<gidx_t>& quad_global_indices) {
         // CALL C-STYLE OPERATOR
             
         size_t nb_nodes = global_indices.size();
         size_t nb_tris  = tri_global_indices.size();
         size_t nb_quads = quad_global_indices.size();
         // assert all related vectors have expected size.
         return operator()(nb_nodes, lons.data(), lats.data(),
                           ghosts.data(), global_indices.data(),
                           remote_indices.data(), partitions.data(),
                           nb_tris,
                           reinterpret_cast<const double*>(tri_boundary_nodes.data()),
                           tri_global_indices.data(),
                           nb_quads,
                           reinterpret_cast<const double*>(quad_boundary_nodes.data()),
                           quad_global_indices.data());
    }
};
```

fmahebert:
Thanks @wdeconinck — these suggestions all make sense; I'll make the changes.

One clarifying question though about your suggested interface: in the C++ `operator()`, what is the reason to reinterpret the `vector<array<gidx_t>>` as a `double*` when passing to the C-style interface?

fmahebert:
Changes made (but changing the reinterpret_cast from `double*` to `gidx_t*`), and tested in the fv3-jedi testbed.

wdeconinck:
> Thanks @wdeconinck — these suggestions all make sense; I'll make the changes.
> 
> One clarifying question though about your suggested interface: in the C++ `operator()`, what is the reason to reinterpret the `vector<array<gidx_t>>` as a `double*` when passing to the C-style interface?
> 
> Edit to clarify: I would expect reinterpreting as a `gidx_t*` as this would better match the intput and output argument types... could this be what you intended?

Indeed, just mistake on my part :)

wdeconinck (changes requested):
Thanks @fmahebert looks great, except I'd like it in the atlas::mesh namespace please.

fmahebert:
Done!

wdeconinck (approved):

PR #137, opened 2023-05-11
Updates to ArrayForEach

wdeconinck:
Change 1
---------

It is best to use the execution::sequential_policy as the default. It has least surprises, and also matches the STL default for its algorithms.

Change 2
----------
It is expected that when we develop a `field::for_each_column` functionality with one-dimensional slices, we expect to be able to pass function arguments in Function (lambda) as:
- `array::LocalView<double, 1>` (read/write)
- `array::LocalView<double,1>&&` (read/write)
- `array::LocalView<const double, 1>` (read-only)
- `array::LocalView<const double,1>&&` (read-only)
- `const array::LocalView<double, 1>&`(read-only)
- `const array::LocalView<const double, 1>&` (read-only)

What is agreeable to not work is:
- `array::LocalView<double,1>&`
- `array::LocalView<const double, 1>&`

In order for this the first list of arguments to be possible, we pass the views to the Function via `std::move`. It makes it impossible to use the second list of arguments.

Also needed to revert the `ArrayViewTuple` template back to `std::tuple<ArrayView...>` for this to work.

With this change, the test_array_foreach requires changes to the slices passed to the lambdas any of following changes can be applied:
1. `auto& slice` -> `auto&& slice` (read/write)
2. `auto& slice` -> `auto slice` (read/write)
3. `auto& slice` -> `const auto& slice` (read-only)

I applied option 1.

Please review @odlomax

### DIFF
--- a/src/atlas/array/helpers/ArrayForEach.h
+++ b/src/atlas/array/helpers/ArrayForEach.h
@@ -15,6 +15,7 @@
 #include "atlas/array/Range.h"
 #include "atlas/array/helpers/ArraySlicer.h"
 #include "atlas/parallel/omp/omp.h"
+#include "atlas/runtime/Exception.h"
 #include "atlas/util/Config.h"
 
 namespace atlas {
@@ -51,10 +52,14 @@ template <>
 constexpr std::string_view policy_name<parallel_unsequenced_policy>() {
   return "parallel_unsequenced_policy";
 };
+template <>
+constexpr std::string_view policy_name<parallel_policy>() {
+  return "parallel_policy";
+};
 
 template <typename execution_policy>
 constexpr std::string_view policy_name(execution_policy) {
-  return policy_name<execution_policy>();
+  return policy_name<std::decay_t<execution_policy>>();
 }
 
 // Type check for execution policy (Not in C++ standard)
@@ -66,6 +71,29 @@ constexpr auto is_execution_policy() { return
   std::is_same_v<execution_policy, unsequenced_policy>;
 }
 
+template <typename ExecutionPolicy>
+constexpr auto demote_policy() {
+  if constexpr(std::is_same_v<ExecutionPolicy, parallel_unsequenced_policy>) {
+    return unseq;
+  }
+  else if constexpr (std::is_same_v<ExecutionPolicy, parallel_policy>) {
+    return seq;
+  }
+  else {
+    return ExecutionPolicy{};
+  }
+}
+
+template <typename execution_policy>
+constexpr auto is_omp_policy() { return
+  std::is_same_v<execution_policy, parallel_policy> ||
+  std::is_same_v<execution_policy, parallel_unsequenced_policy>;
+}
+
+
+template<typename ExecutionPolicy>
+using demote_policy_t = decltype(demote_policy<ExecutionPolicy>());
+
 }  // namespace execution
 
 namespace option {
@@ -74,12 +102,12 @@ namespace option {
 
 template <typename T>
 util::Config execution_policy() {
-  return util::Config("execution_policy", execution::policy_name<T>());
+  return util::Config("execution_policy", execution::policy_name<std::decay_t<T>>());
 }
 
 template <typename T>
 util::Config execution_policy(T) {
-  return execution_policy<T>();
+  return execution_policy<std::decay_t<T>>();
 }
 
 } // namespace option
@@ -89,26 +117,22 @@ namespace helpers {
 
 namespace detail {
 
-template <typename>
-struct IsTupleImpl : std::false_type {};
+struct NoMask {
+  template <typename... Args>
+  constexpr bool operator()(Args...) const { return 0; }
+};
 
-template <typename... Args>
-struct IsTupleImpl<std::tuple<Args...>> : std::true_type {};
-
-template <typename Tuple>
-constexpr auto is_tuple() {
-  return IsTupleImpl<Tuple>::value;
-}
+inline constexpr NoMask no_mask;
 
 template <typename... Ts, typename T>
 constexpr auto tuplePushBack(const std::tuple<Ts...>& tuple, T value) {
   return std::tuple_cat(tuple, std::make_tuple(value));
 }
 
-template <typename Policy, typename Functor>
+template <typename ExecutionPolicy, typename Functor>
 void forEach(idx_t idxMax, const Functor& functor) {
 
-  if constexpr(std::is_same_v<Policy, execution::parallel_unsequenced_policy>) {
+  if constexpr(execution::is_omp_policy<ExecutionPolicy>()) {
     atlas_omp_parallel_for(auto idx = idx_t{}; idx < idxMax; ++idx) {
       functor(idx);
     }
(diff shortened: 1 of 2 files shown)


src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -245,89 +272,89 @@ struct ArrayForEach {
...
+      else if (execution_policy == policy_name(seq)) {
         execute(seq);
       }
+      else {
+        throw_Exception("Unrecognized execution policy "+execution_policy, Here());
+      }
     }
     else {
-      execute(par_unseq);
+      execute(seq);
```
odlomax, line 310: I agree, sequenced default makes sense.

Possible oversight on my behalf, but is it worth changing line 116 to: ` if constexpr(std::is_same_v<Policy, execution::parallel_policy> || std::is_same_v<Policy, execution::parallel_unsequenced_policy>)` ?

I think OpenMP parallel threads support messy things like `new` and `delete`, and are therefore better described by [`par`](https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t).

odlomax (changes requested):
I like the changes -- nobody likes non-const lvalue refs!

Just a couple of minor suggestions.

src/atlas/array/helpers/ArrayForEach.h
odlomax: Reading up a bit further, It would make sense to disable the `apply` interfaces which have a mask argument when the policy is `unseq` or `par_unseq`. The presence of a mask will almost certainly ruin the chances of vectorisation.

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -188,6 +194,12 @@ struct ArrayForEachImpl<ExecutionPolicy, Dim, ItrDim, ItrDims...> {
               tuplePushBack(slicerArgs, idx),
               tuplePushBack(maskArgs, idx));
         }
+        else if constexpr(std::is_same_v<ExecutionPolicy, execution::parallel_policy>) {
```
odlomax, line 197: Perhaps rather than the three `if constexptr, else if constexpr, else` blocks, we could just define a `constexpr auto demote_policy(execution_policy)` function that manages the policy changes.

(I know I put two of them there in the first place ☹)

wdeconinck: Done. Very good idea!

src/atlas/array/helpers/ArrayForEach.h
```diff
@@ -245,89 +272,89 @@ struct ArrayForEach {
...
-  ///         sequential (row-major) order.
+  ///         "sequenced_policy" (default). All loops are then executed in sequential
+  ///         (row-major) order.
+  ///         With "execution_policy" = "parallel_unsequenced" the first loop is executed
+  ///         using OpenMP. The remaining loops are executed in serial.
   ///         Note: The lowest ArrayView.rank() must be greater than or equal
   ///         to the highest dim in ItrDims. TODO: static checking for this.
-  template <typename ArrayViewTuple, typename Mask, typename Function,
-            typename = std::enable_if_t<detail::is_tuple<ArrayViewTuple>()>>
+  template <typename... ArrayView, typename Mask, typename Function>
```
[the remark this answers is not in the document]
wdeconinck: Done.

src/atlas/array/helpers/ArrayForEach.h
wdeconinck: Not 100% sure if we cannot have a mask with unseq though. Compilers can sometimes hoist this thing out for vectorisation. To be further investigated, better another time.

wdeconinck: I implemented the NoMask constexpr struct. It already avoids some std::make_tuple generation of maskArgs, so possible performance gain, who knows.

odlomax: > Not 100% sure if we cannot have a mask with unseq though. Compilers can sometimes hoist this thing out for vectorisation. To be further investigated, better another time.

On reflection, the chances that I'm overthinking this are high!

odlomax (approved):

DJDavies2:
I get a compile failure with this when compiling against oops:

In file included from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/array/ArrayView.h:18,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/functionspace/CubedSphereColumns.h:13,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/functionspace.h:16,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/oops/src/oops/base/GeometryData.h:15,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/oops/src/oops/base/Geometry.h:23,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/oops/src/oops/assimilation/ControlVariable.h:20,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/oops/src/oops/assimilation/ControlIncrement.h:21,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/oops/src/oops/assimilation/MinimizerUtils.h:16,
                 from /home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/oops/src/oops/assimilation/MinimizerUtils.cc:8:
/home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/array/native/NativeArrayView.h:138:66: error: ‘is_convertible_v’ is not a member of ‘std’; did you mean ‘is_convertible’?
  138 |     template <typename ValueTp, typename = std::enable_if_t<std::is_convertible_v<ValueTp*,value_type*>>>
      |                                                                  ^~~~~~~~~~~~~~~~
      |                                                                  is_convertible
/home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/array/native/NativeArrayView.h:138:66: error: ‘is_convertible_v’ is not a member of ‘std’; did you mean ‘is_convertible’?
  138 |     template <typename ValueTp, typename = std::enable_if_t<std::is_convertible_v<ValueTp*,value_type*>>>
      |                                                                  ^~~~~~~~~~~~~~~~
      |                                                                  is_convertible
/home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/array/native/NativeArrayView.h:138:102: error: template argument 1 is invalid
  138 |     template <typename ValueTp, typename = std::enable_if_t<std::is_convertible_v<ValueTp*,value_type*>>>
      |                                                                                                      ^
/home/h01/frwd/cylc-run/atlas-137/share/mo-bundle/atlas/src/atlas/array/native/NativeArrayView.h:138:105: error: expected unqualified-id before ‘>’ token
  138 |     template <typename ValueTp, typename = std::enable_if_t<std::is_convertible_v<ValueTp*,value_type*>>>
      |                                                                                                         ^
gmake[2]: *** [oops/src/CMakeFiles/oops.dir/oops/assimilation/MinimizerUtils.cc.o] Error 1
gmake[1]: *** [oops/src/CMakeFiles/oops.dir/all] Error 2

Is that going to cause a problem?

wdeconinck:
Thanks for checking, @DJDavies2 which compiler is this?

DJDavies2:
This is for gnu (9.3.0).

wdeconinck:
Could I see the full compile line with all compile flags?

wdeconinck:
It seemed like the -std=c++17 flag was not automatically added.
I have modified Atlas to make the flag public. Could you try again now?

PR #140, opened 2023-05-17
Enable latitude normalisation in KDTree coordinate transform

fmahebert:
Depends on https://github.com/ecmwf/eckit/pull/68

Similar changes could be made in many places where `convertSphericalToCartesian` is called, but I wanted to start modest: this is the call that prevents an atlas KDTree from being constructed with an atlas Gaussian grid + structured mesh halo points.


fmahebert:
Thanks @wdeconinck !

PR #141, opened 2023-05-18
Add MPI on/off build configuration to matrix

figi44:
Add build configurations to test atlas and eckit with/without MPI.


figi44:
Need to wait until runners have MPI installed.

wdeconinck:
Does the default have MPI enabled?

figi44:
> Does the default have MPI enabled?

Renamed it to `mpi_on` and `mpi_off`

PR #146, opened 2023-06-07
Interpolate from cubedsphere mesh to structured columns

odlomax:
Hi @wdeconinck,

I've been working on a test to demonstrate functionality that will help @MayeulDestouches with some of his SABER work. It demonstrates how to interpolate from an arbitrarily partitioned CubedSphere mesh (and possibly others) to a StructuredColumns functionspace. This is historically quite tricky, due to the constraints on placed on StructuredColumns. I basically cheat and use a couple of intermediate PointCloud functionspaces to sort out all the redistribution stuff.

One of the things I needed to write was this lambda:

```
    const auto makePointCloud = [](const Grid& grid, const grid::Partitioner partitioner) {

        const auto distribution = grid::Distribution(grid, partitioner);

        auto lonLats = std::vector<PointXY>{};
        auto idx = gidx_t{0};
        for (const auto& lonLat : grid.lonlat()) {
            if (distribution.partition(idx++) == mpi::rank()) {
                lonLats.emplace_back(lonLat.data());
            }
        }
        return functionspace::PointCloud(lonLats);
    };
 ```
 
 Seems to me like there should be a constructor in PointCloud for this. Perhaps two: a Partitioner version and a Distribution version.
 
 Do you agree? If so, it's pretty easy for me to add them.

### CODE REVIEW

File: src/tests/interpolation/test_interpolation_cubedsphere.cc [hunk 1]
```diff
@@ -383,6 +380,81 @@ CASE("cubedsphere_wind_interpolation") {
...
+        auto view         = array::make_view<double, 1>(sourceField);
+        for (idx_t i = 0; i < fixture.sourceFunctionSpace_.size(); ++i) {
+            if (!ghost(i)) {
+                view(i) = util::function::vortex_rollup(lonlat(i, LON), lonlat(i, LAT), 1.);
+            }
+        }
+    }
+    sourceField.haloExchange();
+
+    // Interpolate from source field to targetCubedSpherePointCloud field.
```
odlomax, line 428: @ytremolet @fmahebert 

FYI, this is how you can do a global interpolation with Atlas. It's even simpler if it's a `NodeColumns` to `NodeColumns` interpolation as you don't have to make the intermediate `PointCloud` functionspaces.

odlomax:
Hi @wdeconinck,

This PR fixes a crash that @mayeuldestouches was experiencing when using the "cubedsphere" matching mesh partitioner in SABER. @mayeuldestouches, could you add a link to the issue here.

Cheers!

wdeconinck:
@odlomax thanks for this. The only problem seems to have been the "listSize"?

I agree with a PointCloud constructor as you proposed as well, this could be in a following PR.

odlomax:
> @odlomax thanks for this. The only problem seems to have been the "listSize"?
> 
> I agree with a PointCloud constructor as you proposed as well, this could be in a following PR.

That's correct. Ideally, I'd like a cubed-sphere mesh generator that creates non-disjoint meshes. Then we could use the standard matching mesh tools. A dream for Cubed-Sphere 2.0 perhaps...

wdeconinck (approved):

PR #154, opened 2023-07-18
Fix triangulations for stripack and qhull

fmahebert:
Before commit 0c1fdd66, the `local_indices.clear()` was called _before_ the work of adding the triangles. Commit 0c1fdd66 introduced the `add_triangles` lambda and this moved the code so the clear is called _after_ adding the triangles. 

This change in the order of execution affects the lambda's check of `local_indices.size()` and subsequent logic, producing bad meshes:
CS_tile_stripack

This PR restores the original order, and fixes the meshes:
CS_tile_stripack_orig

### DIFF
--- a/src/atlas/mesh/actions/BuildConvexHull3D.cc
+++ b/src/atlas/mesh/actions/BuildConvexHull3D.cc
@@ -258,6 +258,7 @@ void BuildConvexHull3D::operator()(Mesh& mesh) const {
 
 
     if( local_index.size() == mesh.nodes().size() or local_index.empty() ) {
+        local_index.clear();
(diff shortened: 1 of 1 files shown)


wdeconinck (approved):
Thanks @fmahebert for testing and fixing!

wdeconinck:
@fmahebert, do I understand that you have now tested this and are happy? And can I assume that you are also happy with qhull as backend instead of CGAL/ stripack, in terms of the resulting mesh?

fmahebert:
@wdeconinck Yes — I'm happy with the meshes from this branch and with the qhull backend.

PR #155, opened 2023-07-24
Fix undefined reference due to CGAL dependency

matthewrmshin:
When atlas is built and installed as a standalone package, I would get undefined reference (for symbols associated with the CGAL dependency, e.g., symbols from libgmp, etc) when a subsequent package needs to link to `libatlas.so`. The fix seems to solve the issue, by adding the `CGAL::CGAL` target (if available) to the list of public libs dependency for `libatlas.so`.

### DIFF
--- a/src/atlas/CMakeLists.txt
+++ b/src/atlas/CMakeLists.txt
@@ -937,6 +937,7 @@ ecbuild_add_library( TARGET atlas
     eckit_mpi
     eckit_option
     atlas_io
+    $<TARGET_NAME_IF_EXISTS:CGAL::CGAL>
     $<${atlas_HAVE_EIGEN}:Eigen3::Eigen>
     $<${atlas_HAVE_OMP_CXX}:OpenMP::OpenMP_CXX>
     $<${atlas_HAVE_GRIDTOOLS_STORAGE}:GridTools::gridtools>


MarekWlasak (approved):

matthewrmshin:
@wdeconinck @tlmquintino  please review.

wdeconinck:
This is strange. We have never needed this, and CGAL is a private dependency that should not be public.
Which platform etc. do you encounter this with?
Can it be reproduced with another platform as well?

matthewrmshin:
Hi @wdeconinck I first encounter this while building atlas with atlas-orca as standalone packages on MO's Cray XC40 using a GNU compiler configuration. It would be some linker commands in atlas-orca that would fail with undefined references to some symbols such as `__gmpn_copyi` in `libatlas.so`.

matthewrmshin:
@wdeconinck Thanks for looking at this. I'll try to reproduce the original issue again in my environment. I'll report back when done. Please bear with me in the mean time.

matthewrmshin:
I am pursuing a different method to build and install atlas as a package. The problem here does not appear to be repeatable in my new method, so I'll close this PR for now. Thanks for looking into this.

PR #159, opened 2023-08-30
Register and use non-default MPI communicators in Atlas objects

wdeconinck:
# The problem we want to solve:
Atlas components currently rely on the default communicator registered in eckit and accessed as `eckit::mpi::comm()`.
A Mesh or FunctionSpace which is then generated usually relies on the default communicator as present at its construction time.
When the default communicator is then changed outside of the Atlas context, and e.g. a halo exchange is performed on fields on these meshes or function spaces, then the results are unpredictable and most likely wrong, or the program will just crash or hang.
Alternatively there may be a use case for atlas in different model components with each different MPI communicators. This naturally does not work. A current solution for this would be to context-switch the eckit default communicator to the correct one before each invocation. This is tedious, error-prone, and ugly boiler-plate code.

# The solution
The MPI communicators are stored internally in eckit in a `map<name,Comm>`. Communicators can be inserted and deleted in this map. This mechanism could allow to define model, or mesh/functionspace-specific communicators, which are referable by name.
Without changing Atlas API's we can then pass these communicators as configuration option "mpi_comm", referring to the name registered in eckit. The Mesh or FunctionSpace must then upon construction inspect for this configuration option and keep track of the chosen communicator. If the option is not set the present default communicator is used and kept track of (e.g. "world").

# Plan

- Add `atlas::mpi::comm(std::string_view)` method to access the Comm
- Add mechanism to push/pop the default Comm to cleanly change the default communicator in a certain scope. This is useful to port routines incrementally to use a custom communicator instead.
- Register mpi_comm via configuration in several data structure related components: MeshGenerator, Mesh, FunctionSpace, MatchingPartitioner
- Register mpi_comm via configuration in parallel communication patterns: HaloExchange, GatherScatter, Checksum, Redistribution
- Add tests to verify the functionality

Objects that need to be adapted

- [x] Mesh
- [x] MeshBuilder
- [x] StructuredMeshGenerator
- [x] DelaunayMeshGenerator 
- [x] CubedSphereMeshGenerator
- [x] CubedSphereDualMeshGenerator
- [x] HealpixMeshGenerator
- [x] RegularMeshGenerator 
- [x] PartitionGraph
- [x] PartitionPolygons
- [x] Partitioner
- [x] MatchingPartitioner 
- [x] parallel::HaloExchange
- [x] parallel::GatherScatter
- [x] parallel::Checksum
- [x] functionspace::NodeColumns
- [x] functionspace::CellColumns
- [x] functionspace::EdgeColumns
- [x] functionspace::StructuredColumns
- [x] functionspace::BlockStructuredColumns
- [x] functionspace::CubedSphere...
- [x] functionspace::PointCloud
- [x] Redistribution
- [x] output::Gmsh
- [x] mesh::action::BuildParallelFields
- [x] mesh::action::BuildPeriodicBoundary
- [x] mesh::action::BuildEdges
- [x] mesh::action::BuildHalo

Cannot be done:

- TransPartitioner (underlying ectrans only works with a single communicator)
-  functionspace::Spectral (underlying ectrans only works with a single communicator)

Follow up PR, possibly

- ... other mesh actions ...
- Verify Interpolation, likely working

### DIFF
--- a/src/atlas/functionspace/CellColumns.cc
+++ b/src/atlas/functionspace/CellColumns.cc
@@ -95,7 +95,8 @@ private:
 
     static value_type* create(const Mesh& mesh) {
         value_type* value = new value_type();
-        value->setup(array::make_view<int, 1>(mesh.cells().partition()).data(),
+        value->setup(mesh.mpi_comm(),
+                     array::make_view<int, 1>(mesh.cells().partition()).data(),
                      array::make_view<idx_t, 1>(mesh.cells().remote_index()).data(), REMOTE_IDX_BASE,
                      mesh.cells().size());
         return value;
@@ -129,7 +130,8 @@ private:
 
     static value_type* create(const Mesh& mesh) {
         value_type* value = new value_type();
-        value->setup(array::make_view<int, 1>(mesh.cells().partition()).data(),
+        value->setup(mesh.mpi_comm(),
+                     array::make_view<int, 1>(mesh.cells().partition()).data(),
                      array::make_view<idx_t, 1>(mesh.cells().remote_index()).data(), REMOTE_IDX_BASE,
                      array::make_view<gidx_t, 1>(mesh.cells().global_index()).data(), mesh.cells().size());
         return value;
@@ -193,7 +195,6 @@ void CellColumns::set_field_metadata(const eckit::Configuration& config, Field& 
 }
 
 idx_t CellColumns::config_size(const eckit::Configuration& config) const {
-    const idx_t rank = mpi::rank();
     idx_t size       = nb_cells();
     bool global(false);
     if (config.get("global", global)) {
@@ -201,6 +202,7 @@ idx_t CellColumns::config_size(const eckit::Configuration& config) const {
             idx_t owner(0);
             config.get("owner", owner);
             idx_t _nb_cells_global(nb_cells_global());
+            const idx_t rank = mpi::comm(mpi_comm()).rank();
             size = (rank == owner ? _nb_cells_global : 0);
         }
     }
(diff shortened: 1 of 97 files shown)


fmahebert:
I've been testing this branch for JEDI applications, and it's working well. The interface is unobtrusive and the functionality is as expected.

To be more specific, I'm able to construct multiple NodeColumns FunctionSpaces on a split MPI communicator (one NodeColumns per subcommunicator), and our tests are working as expected. I've tested this for meshes imported into atlas via the MeshBuilder or generated via the DelaunayMeshGenerator. On develop, these tests had been failing with segfaults or MPI hangs.

Thanks @wdeconinck !

fmahebert:
Hi @wdeconinck — on the JEDI side we are seeing two side-effects of this PR (for the small number of us who work with develop atlas and not a tagged release):
- the interface change `mesh.nb_partitions` -> `mesh.nb_parts`
- some runtime errors; have not yet fully investigated.

At the moment we mostly need to investigate more. Perhaps the runtime errors are on our side, or perhaps we've found a defect in this PR? Also, would you consider adding a version bump associated with this PR so that we can more gracefully transition between over from `mesh.nb_partitions` -> `mesh.nb_parts`?

This post is mostly as an FYI, I'm not asking for any changes right now. First need to learn more...

wdeconinck:
Hi @fmahebert, thanks for the feedback. This will be released as part of 0.35.0 I will bump the version.

wdeconinck:
This is now done. The version 0.35.0 will be tagged tomorrow or day after latest.
If it appears other things needed, we can create a 0.35.1 hotfix.

fmahebert:
Hi @wdeconinck — thanks for making the tag and the deprecated transition. This will ease things on our side.

All the other bugs I've found have been cases of atlas becoming more robust and crashing when we wrote bad code. It's clear the new design is more robust. At the same time, the increased consistency across different FunctionSpaces will be really helpful in designing code. Very nice!

(As an example, we had a case of creating a PointCloud with a Grid on rank 0 and an empty list of PointXY on rank 1+... this worked (apparently?) on atlas 0.34 but with 0.35 atlas is crashing. The fix, clearly, is to use a Partitioner to do the assignment of the grid onto rank 0. A good mistake to identify and address.)

PR #161, opened 2023-11-13
Feature/fixes

pmaciel:
Fix compilation warning on handling interpolation matrices (MatrixCache is not fully implemeting the "rule of 5" so I've removed the triggering constructor, the others are sufficient)

Fix a logic error in the generic redistribution method, on detecting indices uniqueness.


wdeconinck (approved):
Thanks for discovering bugs and fixing warnings!

PR #162, opened 2023-11-13
add a getter method for the interpolation matrix

mo-lormi:
## Description
<br>

This PR is intended to implement a 'getter' method that allows the access to the _interpolation matrix_.

A new procedure to perform the regridding from high to low resolution for the CS-LFR grid has been implemented. The procedure uses the Atlas library and in particular the classes `Interpolation` and `Method`.
Currently, the interpolation matrix defined in Atlas is hidden. The changes proposed in this PR will makes the interpolation matrix visible.
<br><br>

All the tests pass.
<br><br>

### DIFF
--- a/src/atlas/interpolation/Interpolation.cc
+++ b/src/atlas/interpolation/Interpolation.cc
@@ -99,6 +99,10 @@ void Interpolation::print(std::ostream& out) const {
     get()->print(out);
 }
 
+const Interpolation::Matrix& Interpolation::matrix() const {
+    return get()->matrix();
+}
+
 const FunctionSpace& Interpolation::source() const {
     return get()->source();
 }
--- a/src/atlas/interpolation/Interpolation.h
+++ b/src/atlas/interpolation/Interpolation.h
@@ -12,6 +12,7 @@
 
 #include "atlas/interpolation/method/Method.h"
 #include "atlas/library/config.h"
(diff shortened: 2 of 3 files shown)


wdeconinck:
Hi @mo-lormi there is in fact already a way. The reason is that some interpolation methods are matrix-free.
The procedure would be:
```
auto matrix_cache = atlas::interpolation::MatrixCache(interpolator);
auto matrix = matrix_cache.matrix();
```
For more involved use and example, see https://github.com/ecmwf/atlas/blob/develop/src/tests/interpolation/test_interpolation_finite_element_cached.cc

You can also use this cache to then recreate a new interpolator if needed.

mo-lormi:
> Hi @mo-lormi there is in fact already a way. The reason is that some interpolation methods are matrix-free. The procedure would be:
> 
> ```
> auto matrix_cache = atlas::interpolation::MatrixCache(interpolator);
> auto matrix = matrix_cache.matrix();
> ```
> 
> For more involved use and example, see https://github.com/ecmwf/atlas/blob/develop/src/tests/interpolation/test_interpolation_finite_element_cached.cc
> 
> You can also use this cache to then recreate a new interpolator if needed.

Hi @wdeconinck many thanks for the suggestion. I am going to close this PR. Thanks again ...

PR #163, opened 2023-11-17
Parallel transport generic vector interpolation method.

odlomax:
Hi @wdeconinck,

I'm experimenting with a new method I've cooked up to perform geometry-agnostic vector interpolation. Essentially, it works out the angular difference between North and a great-circle arc connecting a source and a target point, then rotates the source field vector to match the target field vector.

The upshot of this is that we require ~~four interpolation matrices instead of the usual one~~ complex interpolation weights, but the results look promising. I've experimented below, interpolating from a CS-LFR-48 to an O48 (the interpolation method knows nothing about the cubed-sphere).

Here are some figures showing the results so far:

test_field_eastwards
Fig 1: Eastward component of test field on CS-LFR-48 grid

test_field_northwards
Fig 2: Northward component of test field on CS-LFR-48 grid

error_scalar_interp
Fig 3: Interpolation error when mapped on to O48 gird. Vector field components treated as independent scalars.

error_parallel_transport
Fig 4: Interpolation error when mapped on to O48 gird. Source vectors rotated to match target vectors.

tl;dr I've made the error of interpolating winds over the poles go away, and I believe it will work for all matrix-based interpolation methods.

I'll try and get it finished up over the next week or so.


wdeconinck:
With your updates to rely on _Eigen_ is potentially problematic.
Two ways...
1. Revert the change, which ensures that it will always work.
2. Check if compilation without Eigen works now (cmake : `-DENABLE_EIGEN=OFF`), and
    add some `#if ATLAS_HAVE_EIGEN` in _SphericalVector.h_ and _SphericalVector.cc_ to hide Eigen includes and types and then throw exceptions in the _SphericalVector.cc_ when trying to use it. Also the compilation of _SphericalVector_ should not be optional. The `#if ATLAS_HAVE_EIGEN` in code should take care of it.

odlomax:
> Your updates to exclusively rely on _Eigen_ is potentially problematic. Two ways...
> 
> 1. Revert the change, which ensures that it will always work.
> 2. Check if compilation without Eigen works now (cmake : `-DENABLE_EIGEN=OFF`), and
>    add some `#if ATLAS_HAVE_EIGEN` in _SphericalVector.h_ and _SphericalVector.cc_ to hide Eigen includes and types and then throw exceptions in the _SphericalVector.cc_ when trying to use it. Also the compilation of _SphericalVector_ should not be optional. The `#if ATLAS_HAVE_EIGEN` in code should take care of it.

Thanks, @wdeconinck, I've replaced the optional compilation with pre-processor macros in relevant files. It builds and runs properly on my system with `ENABLE_EIGEN` set to both `ON` and `OFF`.

I did some digging in the current eckit sparse matrix. It does indeed prune out zero-valued triples ([see here](https://github.com/ecmwf/eckit/blob/39c435b8209d64822a62349bdb48cae4f42add54/src/eckit/linalg/SparseMatrix.cc#L135)). This means all the real, imaginary and magnitude matrices can have different sparsity. The simplest in-code solution was to use Eigen directly. Minimal code changes would be required if the current eckit `SparseMatrix` is generalised to `SparseMatrixT`.

I'm going to open an issue which details what I think are the to-do items (hopefully) for follow-up PRs.

src/atlas/interpolation/method/MethodFactory.cc
```diff
@@ -47,6 +48,9 @@ void force_link() {
             MethodBuilder<method::GridBoxAverage>();
             MethodBuilder<method::GridBoxMaximum>();
             MethodBuilder<method::CubedSphereBilinear>();
+#if ATLAS_HAVE_EIGEN
```
wdeconinck, line 51: You don't need preprocessor guard here. It should still self-register, but throw a runtime exception when used.

wdeconinck:
> I did some digging in the current eckit sparse matrix. It does indeed prune out zero-valued triples

That does seem like a sensible optimisation, but prevents us from fusing the two loops indeed, and then Eigen matrix solution is appropriate.

src/atlas/interpolation/method/MethodFactory.cc
odlomax: Done. What sorcery is this?!

src/atlas/interpolation/method/MethodFactory.cc
wdeconinck, line 1: It was still good to have it in this file, but not guarded by `#if ATLAS_HAVE_EIGEN` if that makes sense.

odlomax:
Hi @wdeconinck

I uncovered a lovely little feature where eckit and Eigen3 have slightly different sparsities in their CRS format. I've accounted for that now in the fused matrix-multiply loops. I've also ticket off the code-coverage tests in issue #166 

I believe it's ready for review.

Cheers!

pmaciel:
Hi @odlomax , what is the difference you found between Eigen's and eckit's CSR formats? I'm aware of a few (eckit doesn't require sorted column indices, not even uniqueness (which is a problem)) but is there any other difference? Your input might be invaluable here for an upcoming (separate) rewrite of the SparseMatrix.

I could only see a change re. the iterating through the compressed structure.

odlomax:
> Hi @odlomax , what is the difference you found between Eigen's and eckit's CSR formats? I'm aware of a few (eckit doesn't require sorted column indices, not even uniqueness (which is a problem)) but is there any other difference? Your input might be invaluable here for an upcoming (separate) rewrite of the SparseMatrix.
> 
> I could only see a change re. the iterating through the compressed structure.

On further consideration, my statement is not quite correct. The main difference is that Eigen's `SparseMatrix` constructor sorts each compressed row by `j`, whereas eckit does not.

The underlying data structures are the same, though. You can safely use the Eigen's `Map` template on an eckit `SparseMatrix`.

src/atlas/interpolation/method/MethodFactory.cc
wdeconinck: What I meant is what I added to your branch with [a96057e](https://github.com/ecmwf/atlas/pull/163/commits/a96057eee9cd3d5adc462219c90dc5250248e0da)

odlomax: Ah! That looks far safer!

wdeconinck:
intel/2021.4 failing on our HPC with an internal compiler error :(
```
 /usr/local/apps/intel/2021.4.0/compiler/latest/linux/bin/intel64/icpc -Datlas_EXPORTS -I../src -Isrc -I../atlas_io/src -Iatlas_io/src -I/usr/local/apps/fftw/3.3.10/INTEL/2021.4/include -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/fckit/include -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/fckit/module/fckit -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/fckit/include/fckit -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include/eckit -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include/eckit/geometry -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include/eckit/linalg -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include/eckit/maths -I/opt/mpi/openmpi/4.1.1.1/include -I/opt/mpi/openmpi/4.1.1.1/fortran-intel/include -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include/eckit/mpi -I/etc/ecmwf/ssd/ssd1/tmpdirs/***.33879484/install/eckit/include/eckit/option -I/usr/local/apps/eigen/3.4.0/include/eigen3 -O2 -g -DNDEBUG -fPIC -qopenmp -std=gnu++17 -MD -MT src/atlas/CMakeFiles/atlas.dir/interpolation/method/sphericalvector/SphericalVector.cc.o -MF src/atlas/CMakeFiles/atlas.dir/interpolation/method/sphericalvector/SphericalVector.cc.o.d -o src/atlas/CMakeFiles/atlas.dir/interpolation/method/sphericalvector/SphericalVector.cc.o -c ../src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
  ../src/atlas/interpolation/method/sphericalvector/SphericalVector.cc(70): internal error: assertion failed: find_assoc_pragma: pragma not found (il.c, line 25413 in find_assoc_pragma)
  
          functor(it.row(), it.col(), it.value());
          ^
  
  compilation aborted for ../src/atlas/interpolation/method/sphericalvector/SphericalVector.cc (code 4)
  ```

**Comment by wdeconinck** [comment]
Nice cleanup.
Unfortunately still the same compile error with intel/2021.4

**Comment by odlomax** [comment]
> Nice cleanup. Unfortunately still the same compile error with intel/2021.4 We could also disable OpenMP for that compiler and version as that [thread](community.intel.com/t5/Intel-C-Compiler/compiler-2020-1-217-internal-error-assertion-failed-find-assoc/m-p/1386077#M40012) you suggested says it could be fixed with 2022.2

It's certainly worth a try. Should at least help diagnose the problem! 😊

File: src/atlas/interpolation/method/sphericalvector/SphericalVector.cc [hunk 2]
```diff
@@ -40,6 +40,16 @@ MethodBuilder<SphericalVector> __builder("spherical-vector");
...
+// A bug exists in intel versions < intel/2022.2 with OpenMP
+// Intel OneAPI version 2022.2 corresponds to Intel classic (icpc) version 2021.6
+#if defined(__INTEL_COMPILER) && defined(__INTEL_COMPILER_UPDATE)
+#if (__INTEL_COMPILER <= 2021) && (__INTEL_COMPILER_UPDATE < 6)
+#warning Disabling OpenMP to prevent internal compiler error for intel-classic version < 2021.6 (intel-oneapi/2022.2)
+#undef atlas_omp_parallel_for
+#define atlas_omp_parallel_for for
+#endif
+#endif
+
```
odlomax, line 52: Thanks, @wdeconinck. Hit it with macros until the problem goes away! ❤

wdeconinck: The problem indeed went away 😃

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+  const auto sourceView = array::make_view<Value, Rank>(sourceField);
+  auto targetView = array::make_view<Value, Rank>(targetField);
+  targetView.assign(0.);
+
+  const auto horizontalComponent = [](const auto& sourceVars, auto& targetVars,
+                                      const auto& complexWeight) {
+    const auto sourceVector = Complex(sourceVars(0), sourceVars(1));
+    const auto targetVector = complexWeight * sourceVector;
+    targetVars(0) += targetVector.real();
+    targetVars(1) += targetVector.imag();
```
wdeconinck, line 267: Without going into the detailed maths, why do you have to
`targetView.assign(0)`
and then use `+=` for the targetVars ? Can this not be replaced with just `=` ?

odlomax: It's because each target element (named `targetVars` here) has a contribution from multiple source elements (`sourceVars`). If I changed `+=` to `=`, it would set the target element to the last source element on that row of the matrix.

odlomax: I could do the `assign(0.)` within `matrixMultiply` function, but it would have to be on the first line, and therefore be the equivalent operation as it currently is. The `SparseMatrixForEach` is the only function that has knowledge of `i` as a whole row, but it doesn't know what `functor` is doing (it's used in a couple of different contexts).

I suspect details like this will have to be refactored to remove code duplication when the adjoint methods are added. In that context, you don't assign zero to the array you're writing to.

wdeconinck: OK it can stay as-is. We should not attempt premature optimisation now.

wdeconinck (approved):

src/atlas/interpolation/method/sphericalvector/SphericalVector.h
```diff
@@ -0,0 +1,127 @@
...
+#include <memory>
+
+#if ATLAS_HAVE_EIGEN
+#include <Eigen/Sparse>
+#endif
+
+#include "atlas/functionspace/FunctionSpace.h"
+#include "atlas/interpolation/method/Method.h"
+#include "atlas/linalg/sparse.h"
+#include "eckit/config/Configuration.h"
```
pmaciel, line 22: The correct include is "eckit/config/LocalConfiguration.h", and should precede the "atlas/"* includes

src/atlas/interpolation/method/sphericalvector/SphericalVector.h
```diff
@@ -0,0 +1,127 @@
...
+
+  void do_execute(const FieldSet& sourceFieldSet, FieldSet& targetFieldSet,
+                  Metadata& metadata) const override;
+  void do_execute(const Field& sourceField, Field& targetField,
+                  Metadata& metadata) const override;
+
+  void do_execute_adjoint(FieldSet& sourceFieldSet,
+                          const FieldSet& targetFieldSet,
+                          Metadata& metadata) const override {
+    ATLAS_NOTIMPLEMENTED;
```
pmaciel, line 72: I think a better way to express this is in the .cc, so the API doesn't change, and the correct #include is already there, not here (also below).

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+#include "atlas/library/defines.h"
+
+#include <cmath>
+#include <tuple>
+#include <utility>
```
pmaciel, line 12: missing <type_traits>, <iterator>, but remove <cmath>

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+#include "atlas/interpolation/method/sphericalvector/SphericalVector.h"
+#include "atlas/linalg/sparse.h"
+#include "atlas/option/Options.h"
+#include "atlas/parallel/omp/omp.h"
+#include "atlas/runtime/Exception.h"
+#include "atlas/runtime/Trace.h"
+#include "atlas/util/Constants.h"
+#include "atlas/util/Geometry.h"
+
+#include "eckit/linalg/Triplet.h"
```
pmaciel, line 31: (before "atlas/"*)

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+
+#include "atlas/array/ArrayView.h"
+#include "atlas/array/helpers/ArrayForEach.h"
+#include "atlas/array/Range.h"
+#include "atlas/field/Field.h"
+#include "atlas/field/FieldSet.h"
+#include "atlas/interpolation/Cache.h"
+#include "atlas/interpolation/Interpolation.h"
+#include "atlas/interpolation/method/MethodFactory.h"
+#include "atlas/interpolation/method/sphericalvector/SphericalVector.h"
```
pmaciel, line 22: Normally the header specific to the class implementation comes before any other include, so to guarantee the implementation sees the same order of includes from the declarion

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+    const auto targetVector = complexWeight * sourceVector;
+    targetVars(0) += targetVector.real();
+    targetVars(1) += targetVector.imag();
+  };
+
+  if (sourceField.variables() == 2) {
+    matrixMultiply(sourceView, targetView, horizontalComponent,
+                   *complexWeights_);
+    return;
+  } else if (sourceField.variables() == 3) {
```
pmaciel, line 274: "else if" should be "if" instead (because of the "return").

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+#include "atlas/array/helpers/ArrayForEach.h"
+#include "atlas/array/Range.h"
+#include "atlas/field/Field.h"
+#include "atlas/field/FieldSet.h"
+#include "atlas/interpolation/Cache.h"
+#include "atlas/interpolation/Interpolation.h"
+#include "atlas/interpolation/method/MethodFactory.h"
+#include "atlas/interpolation/method/sphericalvector/SphericalVector.h"
+#include "atlas/linalg/sparse.h"
+#include "atlas/option/Options.h"
```
pmaciel, line 24: Remove "atlas/linalg/sparse.h" and "atlas/option/Options.h"

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
```diff
@@ -0,0 +1,296 @@
...
+    Method::do_execute(sourceField, targetField, metadata);
+
+    return;
+  }
+
+  if (target_.size() == 0) {
+    return;
+  }
+
+  ATLAS_ASSERT_MSG(sourceField.variables() == 2 || sourceField.variables() == 3,
```
pmaciel, line 224: Shouldn't this assertion come right at the beggining of the method?

pmaciel (approved):
Hi @odlomax,

as discussed this is a great contribution! As you see, I've only  commented on technical details, because it looks like everything is well thought of. I admit, I think there is a bit overuse of templates to pass functors (when we really only handle the real and complex cases), but it is certainly not an important concern at the moment.

I've already put the "course"  method as part of a revised eckit geometry library. I look forward to have ready the more flexible sparse matrix container and the sparse linear algebra backend too, as we discussed. It will happen in the coming months, and will also simplify much of the workarounds you had to painfuly go through.

I'm already aproving the PR -- this is ready to merge!

odlomax:
> Hi @odlomax,
> 
> as discussed this is a great contribution! As you see, I've only commented on technical details, because it looks like everything is well thought of. I admit, I think there is a bit overuse of templates to pass functors (when we really only handle the real and complex cases), but it is certainly not an important concern at the moment.
> 
> I've already put the "course" method as part of a revised eckit geometry library. I look forward to have ready the more flexible sparse matrix container and the sparse linear algebra backend too, as we discussed. It will happen in the coming months, and will also simplify much of the workarounds you had to painfuly go through.
> 
> I'm already aproving the PR -- this is ready to merge!

Thanks, @pmaciel, @wdeconinck,

I had a lot of fun with this one.

I freely admit I'm a template addict. If only we had C++ 20, I could at least throw a few concepts on them! I need to follow this up with an implementation of the adjoint methods. Shall I push @pmaciel's suggested changes to this PR, or the next one?

pmaciel:
> Thanks, @pmaciel, @wdeconinck,
> 
> I had a lot of fun with this one.
> 
> I freely admit I'm a template addict. If only we had C++ 20, I could at least throw a few concepts on them! I need to follow this up with an implementation of the adjoint methods. Shall I push @pmaciel's suggested changes to this PR, or the next one?
> 
> 

Since they're only technical would prefer that it's merged with the suggested changes :-)

src/atlas/interpolation/method/sphericalvector/SphericalVector.h
odlomax: Done. Also moved to the .cc along with constructor definition.

src/atlas/interpolation/method/sphericalvector/SphericalVector.h
odlomax: Pure laziness on my part! Done.

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
odlomax: Done.

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
odlomax: Done.

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
odlomax: Done.

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
odlomax: Clearly I've lazily attempted to refactor at some point. Done.

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
odlomax: Done.

src/atlas/interpolation/method/sphericalvector/SphericalVector.cc
odlomax: I'd also have to check that the field was a vector type. By this point in the function, it definitely is.

odlomax:
> > Thanks, @pmaciel, @wdeconinck,
> > I had a lot of fun with this one.
> > I freely admit I'm a template addict. If only we had C++ 20, I could at least throw a few concepts on them! I need to follow this up with an implementation of the adjoint methods. Shall I push @pmaciel's suggested changes to this PR, or the next one?
> 
> Since they're only technical would prefer that it's merged with the suggested changes :-)

It is done! Thanks, @pmaciel 🙂

wdeconinck:
Thanks @odlomax for addressing further review. I will merge today when all tests have passed. This will be super useful.

odlomax:
> Thanks @odlomax for addressing further review. I will merge today when all tests have passed. This will be super useful.

Again, thanks @wdeconinck and @pmaciel. Always a pleasure.

PR #165, opened 2023-11-21
Feature/fortran api init functions

sbrdar:
Add Fortran API for initial functions util/function/MDPI_functions


src/tests/util/fctest_functions.F90
```diff
@@ -0,0 +1,73 @@
...
+
+! -----------------------------------------------------------------------------
+
+TEST( test_atlas_Functions )
+  real(c_double) :: val
+  val = MDPI_sinusoid(1._c_double, 1._c_double)
+  val = MDPI_harmonic(1._c_double, 1._c_double)
+  val = MDPI_vortex(1._c_double, 1._c_double)
+  val = MDPI_gulfstream(1._c_double, 1._c_double)
+END_TEST
```
wdeconinck, line 47: Could you add a few more lines to check values are correct up to some tolerance?
That is to make sure that the returned value is not by mistake unassigned.

src/tests/util/fctest_functions.F90
```diff
@@ -0,0 +1,73 @@
...
+
+TEST( test_atlas_Functions_vector )
+  real(c_double), dimension(3) :: val, lon, lat
+  lon = [ 1._c_double, 1._c_double, 1._c_double ]
+  lat = [ 1._c_double, 1._c_double, 1._c_double ]
+  val = MDPI_sinusoid(lon, lat)
+  val = MDPI_harmonic(lon, lat)
+  val = MDPI_vortex(lon, lat)
+  val = MDPI_gulfstream(lon, lat)
+END_TEST
```
wdeconinck, line 57: Could you also add here a few more value checks, and make the lon and lat values different (e.g. 1, 2, 3)

src/tests/util/fctest_functions.F90
```diff
@@ -0,0 +1,73 @@
...
+
+TEST( test_initialise_field )
+  type(atlas_Field) :: field_xy, field_val
+  real(c_double), dimension(:,:), pointer :: field_xy_v
+  real(c_double), dimension(:), pointer :: field_val_v
+  field_xy = atlas_Field(kind=atlas_real(c_double), shape=[3,3])
```
wdeconinck, line 63:
```suggestion
  field_xy = atlas_Field(kind=atlas_real(c_double), shape=[2,3])
```
The inner dimension should be 2, i.e. 2 coordinates per point.

wdeconinck (changes requested):
Thanks @sbrdar that will be useful for testing algorithms. Please address following issues.

src/tests/util/fctest_functions.F90
sbrdar: Thanks for the comment @wdeconinck. I have addressed the issues.

wdeconinck (approved):
Thanks @sbrdar for addressing all comments!

PR #169, opened 2024-01-19
Add workflow to build and deploy documentation

figi44:
Adds a workflow which builds documentation and publishes it to sites.ecmwf.int


wdeconinck:
This is great!
Could the location of doc-config.yml be configurable, and be within the .github directory?

PR #168, opened 2024-01-08
Refactoring of interpolation::method::SphericalVector and implementation of adjoint methods.

odlomax:
Following on from [PR #163](https://github.com/ecmwf/atlas/pull/163), this PR primarily adds and tests the adjoint methods for the `interpolation::method::SphericalVector` class.

Several refactors were also applied to reduce complexity and code duplication:

- The matrix multiplication code was removed from `SphericalVector.cc` and given it's own class `ComplexMatrixMultiply`
- A SparseMatrix class was added that wraps the Eigen Sparse matrix into something that looks more like `eckit::linalg::SparseMatrix` with a `Value` template. This also confines all the nasty macros to one place.
- `Types.h` was added, containing all the main type definitions required for the `SphericalVector` class.
- `ArrayForEachDim` was added to `array::helpers::ArrayForEach`. This allows the iteration dimensions to be set using an `std::integer_sequence`.

Further testing was added

- Test added for `ArrayForEachDim` in `ArrayForEach` tests.
- Adjoint methods added to all interpolation tests.
- StructuredColumns to CubedSphere interpolation test added. Figures showing performance are given below. Note: `StructuredColumns` `FixupHaloForVectors` methods interferes with `SphericalVector`. You *must* make sure a halo exchange isn't called on the source field for it to work.

Notes:
Apologies for the strange commit history, I was trying to debug the MacOS build without a Mac.
Closes #166 

### Figures

O48 Structured Columns to Cubed Sphere interpolation error when using `SphericalVector` to treat vector fields (same setup as  [PR #163](https://github.com/ecmwf/atlas/pull/163)).

sc_to_cs_spherical

O48 Structured Columns to Cubed Sphere interpolation error when using `FixupHaloForVectors` to treat vector fields.
sc_to_cs_spherical_raw

PR #171, opened 2024-01-25
Added arrayForEachDim method to simplify template metaprogramming.

odlomax:
This PR adds the following method to `src/atlas/array/helpers/ArrayForEach.h`:

```
/// brief   Construct ArrayForEach and call apply
///
/// details Construct an ArrayForEach<ItrDims...> using std::integer_sequence
///         <int, ItrDims...>. Remaining arguments are forwarded to apply
///         method.
template <int... ItrDims, typename... Args>
void arrayForEachDim(std::integer_sequence<int, ItrDims...>, Args&&... args) {
  ArrayForEach<ItrDims...>::apply(std::forward<Args>(args)...);
}
```

Relevant tests are added to `src/tests/array/test_array_foreach.cc`.

This feature supports #168 and should be merged beforehand.

### DIFF
--- a/src/atlas/array/helpers/ArrayForEach.h
+++ b/src/atlas/array/helpers/ArrayForEach.h
@@ -1,5 +1,5 @@
 /*
- * (C) Crown Copyright 2023 Met Office
+ * (C) Crown Copyright 2024 Met Office
  *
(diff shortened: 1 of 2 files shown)


wdeconinck (approved):
Great, no comments :)

PR #172, opened 2024-02-27
Bugfix for regional grids if ny > nx

benjaminmenetrier:
I've found a minor bug affecting regional grids if ny > nx.
I guess it is due to a copy/paste error from line 428.


wdeconinck (approved):
Thanks @benjaminmenetrier for spotting this!

benjaminmenetrier:
Thank you @wdeconinck!

PR #175, opened 2024-03-04
Update `SphericalVector` to work with StructuredColumns as source functionspace.

### LINKED ISSUE #174: StructuredColumns `FixupHaloForVectors` method interferes with `SphericalVector` interpolation
### What happened?

`StructuredColumns` contains a struct template [`FixupHaloForVectors`](https://github.com/ecmwf/atlas/blob/4e2a05546bd32ed6762da9fcf69ca4a1978dc72e/src/atlas/functionspace/detail/StructuredColumns.cc#L724) which is applied to fields which have `type` set to `vector` in their metadata. This multiplies field values beyond the poles by `-1` directly after a halo exchange.

The `SphericalVector` interpolation scheme also checks for `type == vector` before applying the vector interpolation method. However, the factor of `-1` applied by `FixupHaloForVectors` corrupts the interpolation at the poles (see below).
fixup_halo_exchange

The work-around, currently used in the `atlas_test_interpolation_spherical_vector` ctest is to make sure that the `dirty` value in the field metadata is set to `false`. This produces a clean interpolation at the poles (see below), but it's clearly not useful in the general case where halo-exchanges are required.
fixup_no_halo_exchange

### What are the steps to reproduce the bug?

The bug can be reproduced by commenting out [this line](https://github.com/ecmwf/atlas/blob/4e2a05546bd32ed6762da9fcf69ca4a1978dc72e/src/tests/interpolation/test_interpolation_spherical_vector.cc#L210) from the interpolation test.

### Version

develop

### Platform (OS and architecture)

All builds

### Relevant log output

_No response_

### Accompanying data

_No response_

### Organisation

Met Office

odlomax:
This draft PR updates the test from PR #168 to include halo exchanges. Due to issue #174, the following code has to be added before `Inderpolation::exectute` is called:

```
  if (functionspace::StructuredColumns(sourceFunctionSpace)) {
    sourceFieldSet[0].metadata().set("type", "scalar");
    sourceFieldSet[0].haloExchange();
    sourceFieldSet[0].metadata().set("type", "vector");
  }
```
For other functionspaces, the type is set to `vector` and  the halo exchange is called within `Inderpolation::exectute`.

If you comment out this block, you'll reproduce the output in #174 .

Once #174 is addressed, this block can be deleted and the PR will be ready for review.

Edit: further additions to this PR will close #174

### DIFF
--- a/src/atlas/util/Geometry.cc
+++ b/src/atlas/util/Geometry.cc
@@ -3,9 +3,9 @@
  *
  * This software is licensed under the terms of the Apache Licence Version 2.0
  * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
- * In applying this licence, ECMWF does not waive the privileges and
- * immGeometryies granted to it by virtue of its status as an intergovernmental
- * organisation nor does it submit to any jurisdiction.
+ * In applying this licence, ECMWF does not waive the privileges and immunities
+ * granted to it by virtue of its status as an intergovernmental organisation
+ * nor does it submit to any jurisdiction.
  */
 
 #include "atlas/util/Geometry.h"
--- a/src/atlas/util/Geometry.cc
+++ b/src/atlas/util/Geometry.cc
@@ -15,8 +15,6 @@
 #include "atlas/library/config.h"
 #include "atlas/runtime/Exception.h"
 #include "atlas/util/Constants.h"
-#include "eckit/geometry/Point2.h"
-#include "eckit/geometry/Point3.h"
 
 namespace atlas {
 
--- a/src/atlas/util/Geometry.cc
+++ b/src/atlas/util/Geometry.cc
@@ -3,21 +3,20 @@
  *
  * This software is licensed under the terms of the Apache Licence Version 2.0
  * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
- * In applying this licence, ECMWF does not waive the privileges and immGeometryies
- * granted to it by virtue of its status as an intergovernmental organisation
- * nor does it submit to any jurisdiction.
+ * In applying this licence, ECMWF does not waive the privileges and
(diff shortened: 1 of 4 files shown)


MarekWlasak:
Is there a better way to sort this out? ... The code above looks a bit hacky.

odlomax:
> Is there a better way to sort this out? ... The code above looks a bit hacky.

"Once https://github.com/ecmwf/atlas/issues/174 is a addressed, this block can be deleted and the PR will be ready for review."

odlomax:
@wdeconinck @MarekWlasak I've commented on a very simple solution in #174  which I believe this PR closes.

I've also added some further tests to make sure sensible values are returned for points located on a pole.

Thanks, all!

src/atlas/util/Geometry.cc
```diff
@@ -45,7 +43,10 @@ void GeometrySphere::xyz2lonlat(const Point3& xyz, Point2& lonlat) const {
 /// @ref     https://en.wikipedia.org/wiki/Great-circle_navigation
 ///
 std::pair<double, double> greatCircleCourse(const Point2& lonLat1,
-                                                   const Point2& lonLat2) {
+                                            const Point2& lonLat2) {
+  if (lonLat1 == lonLat2) {
```
pmaciel, line 47: This should be using bool points_equal(const Point2&, const Point2&)

I have a more consistent way to compare these in an upcoming PR (a refactoring of the geometry library) and I've already included the "course" methods there, but they will need updating to your improved version, when that time comes :-)

wdeconinck: That's correct, that it should use "points_equal". But that already happens. See https://github.com/ecmwf/atlas/blob/develop/src/atlas/util/Point.h#L32-L34.

wdeconinck (approved):
Thanks @odlomax for the fixes and discussions.

PR #179, opened 2024-03-08
Feature/area extraction test

antons-it:
Area extraction unit tests that fails with -np > 2

### DIFF
--- a/src/atlas/functionspace/PointCloud.cc
+++ b/src/atlas/functionspace/PointCloud.cc
@@ -170,7 +170,6 @@ PointCloud::PointCloud(const Grid& grid, const grid::Partitioner& _partitioner, 
 
     if (halo_radius == 0. || nb_partitions_ == 1) {
         idx_t size_halo = size_owned;
-        ATLAS_ASSERT(size_owned > 0);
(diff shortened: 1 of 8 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/atlas?pullRequest=179) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/atlas?pullRequest=179) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/atlas?pullRequest=179) it.</sub>

wdeconinck:
This test is still failing with MPI_SIZE = 120 and with grid O400.
Partition 76 coincidentally ends up with only very few points with which a mesh cannot be generated.
image

A workaround, but NOT the solution, could be choosing different MPI_SIZE (e.g. 128) or moving the area domain to include more points on partition 76 (or none at all).

wdeconinck:
> This test is still failing with MPI_SIZE = 120 and with grid O400.

Other possibility is to use `functionspace::PointCloud` as a target. Then no mesh is attempted to be generated for partition 76

PR #182, opened 2024-03-08
Feature/gpu fields

sbrdar:
Allow CMake to build storage either with GridTools, or Native-CUDA, or simply Native-host 
Add CUDA backend for Atlas Native-CUDA


src/atlas/CMakeLists.txt
```diff
@@ -974,7 +977,7 @@ ecbuild_add_library( TARGET atlas
     $<${atlas_HAVE_EIGEN}:Eigen3::Eigen>
     $<${atlas_HAVE_OMP_CXX}:OpenMP::OpenMP_CXX>
     $<${atlas_HAVE_GRIDTOOLS_STORAGE}:GridTools::gridtools>
-    $<$<AND:${atlas_HAVE_CUDA},${atlas_HAVE_NATIVE_STORAGE}>:CUDA::cudart>
+    $<${atlas_HAVE_NATIVE_STORAGE}:CUDA::cudart>
```
wdeconinck: This should only be the case when CUDA is enabled, so: ```suggestion
    $<${ATLAS_NATIVE_STORAGE_BACKEND_CUDA}:CUDA::cudart>
```

Comment (wdeconinck) [inline src/atlas/CMakeLists.txt]:
```suggestion
    $<${ATLAS_NATIVE_STORAGE_BACKEND_CUDA}:CUDA::cudart>
```
because this should only be the case when CUDA is enabled.

wdeconinck:
Thanks @sbrdar for your contribution to improve the GPU support in Atlas! Much appreciated 🙏

PR #184, opened 2024-03-23
Fix build for configuration setting ATLAS_BITS_LOCAL=64

pmaciel:
I've tried matching the arguments to the signatures types (maybe there should only be one GatherScatter::setup). This seems to have been a requirement put forward for production ("for O8000" is the comment), so as far as I can tell is is urgent.


wdeconinck:
Thanks @pmaciel . This should really become a CI configuration to avoid this from happening in future.

PR #185, opened 2024-03-27
Projection base implementation derivatives performance/encapsulation …

pmaciel:
Projection base implementation derivatives performance/encapsulation improvements (better use of temporaries). This comes from the eckit::geo library.

Mesh adds a missing header atlas/runtime/Exception.h

### DIFF
--- a/src/atlas/projection/detail/MercatorProjection.cc
+++ b/src/atlas/projection/detail/MercatorProjection.cc
@@ -14,7 +14,6 @@
 #include <sstream>
 
 #include "eckit/config/Parametrisation.h"
-#include "eckit/types/FloatCompare.h"
 #include "eckit/utils/Hash.h"
 
 #include "atlas/projection/detail/MercatorProjection.h"
@@ -98,22 +97,18 @@ void MercatorProjectionT<Rotation>::lonlat2xy(double crd[]) const {
     rotation_.unrotate(crd);
 
     // then project
-    if (eckit::types::is_approximately_equal<double>(crd[LAT], 90., 1e-3)) {
+    if (crd[LAT] >= 90. - 1e-3) {
         crd[XX] = false_easting_;
         crd[YY] = std::numeric_limits<double>::infinity();
-        return;
     }
-
-    if (eckit::types::is_approximately_equal<double>(crd[LAT], -90., 1e-3)) {
+    else if (crd[LAT] <= -90. + 1e-3) {
         crd[XX] = false_easting_;
         crd[YY] = -std::numeric_limits<double>::infinity();
-        return;
     }
-
-    crd[XX] = k_radius_ * (D2R(normalise_mercator_(crd[LON]) - lon0_));
-    crd[YY] = k_radius_ * 0.5 * std::log(t(crd[LAT]));
-    crd[XX] += false_easting_;
-    crd[YY] += false_northing_;
+    else {
+        crd[XX] = false_easting_ + k_radius_ * (D2R(normalise_mercator_(crd[LON]) - lon0_));
+        crd[YY] = false_northing_ + k_radius_ * 0.5 * std::log(t(crd[LAT]));
+    }
 }
 
 template <typename Rotation>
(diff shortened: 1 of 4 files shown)


wdeconinck:
Thanks @pmaciel for cleanup and possibly resulting performance improvements. What caused you to tackle this?
At first sight looks OK, but a number of tests are now failing with FE_DIV_BY_ZERO:
  The following tests FAILED:
  	 75 - atlas_test_grid_regional_lambert_conformal_conic_1 (Failed)
  	 76 - atlas_test_grid_regional_lambert_conformal_conic_2 (Failed)
  	 78 - atlas_test_grid_regional_mercator_1 (Failed)
  	 79 - atlas_test_grid_regional_mercator_2 (Failed)
  	 80 - atlas_test_grid_regional_mercator_3 (Failed)

src/atlas/projection/detail/ProjectionImpl.h
```diff
@@ -49,8 +45,7 @@ class ProjectionImpl : public util::Object {
     static const ProjectionImpl* create(const eckit::Parametrisation& p);
     static const ProjectionImpl* create(const std::string& type, const eckit::Parametrisation& p);
 
-    ProjectionImpl()          = default;
-    virtual ~ProjectionImpl() = default;  // destructor should be virtual
```
wdeconinck, line 53: Is there no need for a virtual destructor?

src/atlas/projection/detail/ProjectionImpl.cc
```diff
@@ -53,39 +51,27 @@ struct DerivateBuilder : public ProjectionImpl::DerivateFactory {
...
+    PointLonLat d(PointXY P) const override { return (xy2lonlat(P + H2_) - xy2lonlat(P - H2_)) * invnH_; }
 };
 
 }  // namespace
 
 ProjectionImpl::Derivate::Derivate(const ProjectionImpl& p, PointXY A, PointXY B, double h, double refLongitude):
     projection_(p),
     H_{PointXY::mul(PointXY::normalize(PointXY::sub(B, A)), h)},
-    normH_(PointXY::norm(H_)),
+    invnH_(1. / PointXY::norm(H_)),
```
wdeconinck, line 74: The only place I can think of where FE_DIV_BY_ZERO could originate is here.

pmaciel: I haven't been able to trigger the FPEs yet; what combination of platform/compilers do these FPEs trigger on?

pmaciel:
> Thanks @pmaciel for cleanup and possibly resulting performance improvements. What caused you to tackle this? At first sight looks OK, but a number of tests are now failing with FE_DIV_BY_ZERO: The following tests FAILED: 75 - atlas_test_grid_regional_lambert_conformal_conic_1 (Failed) 76 - atlas_test_grid_regional_lambert_conformal_conic_2 (Failed) 78 - atlas_test_grid_regional_mercator_1 (Failed) 79 - atlas_test_grid_regional_mercator_2 (Failed) 80 - atlas_test_grid_regional_mercator_3 (Failed)

Good question; these are just a set of changes I had here when porting the bounding box calculation to eckit::geo (it is a requirement for earthkit.maps). the changes are only a cleanup/minor performace regarding avoiding so many temporaries of KPoint, and precompute recurring divisions -- which the bisection methods relies on. Functionality should be absolutely identical, so I'm surprised by the FPE issue being triggered (good catch, I have to cover for that for sure).

What's the platform/compiler combination for this triggering?

wdeconinck (approved):
Thanks seems to have fixed it!

src/atlas/projection/detail/MercatorProjection.cc
```diff
@@ -96,6 +98,18 @@ void MercatorProjectionT<Rotation>::lonlat2xy(double crd[]) const {
     rotation_.unrotate(crd);
 
     // then project
+    if (eckit::types::is_approximately_equal<double>(crd[LAT], 90., 1e-3)) {
```
wdeconinck, line 101: To the same and faster effect, could this not be replaced with:
```c++
double constexpr lat_max = 90. - 1.e-3;
if ( crd[LAT] > lat_max ) {
```

pmaciel: Will do

pmaciel: I've simplified the check of polar closeness

PR #191, opened 2024-05-03
implement a procedure to carry out a regridding from high to low resolution (binning)

mo-lormi:
<br>

## Description
<br>

This PR is intended to extend the group of the interpolation operations, in particular to:

* implement a procedure to evaluate the _areas of the grid cells_ (grid type: CS-LFR)
* implement a procedure to carry out a _regridding from high to low resolution (binning)_
* add new tests to the test battery
<br><br>

Rigridding from _high_ to _low_ resolution via _binning_ ...

_method:_ binning
_binning matrix_: **B = N A^T W**
_area weights matrix_: **W**
_interpolation matrix_: **A**
_normalization factor matrix_: **N**
<br><br>

All the tests pass.
<br><br>


mo-lormi:
@wdeconinck , @odlomax  This PR covers the implementation of the regridding procedure (from high to low resolution).
At this stage, I implemented a simple test associated with the procedure to generate data in Gsmh format for data visualization. Further tests will be added.
<br>

odlomax:
Thanks, @mo-lormi 

Could you add some of the outputs to this PR so we can easily have a look?

mo-lormi:
The following plots refer to two fields:
1) _field:_ field_01_s ; _regridding type:_ high to low; _source grid:_ CS-LFR-100 ;
2) _field:_ field_01_t ; _regridding type:_ high to low; _target grid:_ CS-LFR-50 ;

regridding_h2l_field_s
regridding_h2l_field_t

twsearle:
This looks very nice! I am curious as a bystander why you haven't used the interpolation elements. Perhaps if you had it would have worked for any grid? https://github.com/ecmwf/atlas/blob/develop/src/atlas/interpolation/element/Quad2D.h

Maybe out of scope though!

mo-lormi:
<br>

I carried further testing. The procedure also works for a different type of grid (function space: _StructuredColumns_).

The following plots refer to two fields:
1) _field:_ field_02_s ; _regridding type:_ high to low; _source grid:_ O32
2) _field:_ field_02_t ; _regridding type:_ high to low; _target grid:_ O16
regridding_h2l_field02_s
regridding_h2l_field02_t

wdeconinck:
@mo-lormi concerning the "missing" part in the Gmsh plot, is nothing to worry about. That's because the StructuredColumns fields are unlike NodeColumns and don't have extra duplicated nodes at 360 degrees.

mo-lormi:
> @mo-lormi concerning the "missing" part in the Gmsh plot, is nothing to worry about. That's because the StructuredColumns fields are unlike NodeColumns and don't have extra duplicated nodes at 360 degrees.

@wdeconinck thanks for this info ... To take into account this point, I will need to update the procedure.
<br>

src/tests/interpolation/CMakeLists.txt
```diff
@@ -123,11 +121,18 @@ ecbuild_add_test( TARGET atlas_test_interpolation_cubedsphere
...
-  ENVIRONMENT ${ATLAS_TEST_ENVIRONMENT}
+  OMP      4
   CONDITION atlas_HAVE_EIGEN
+  ENVIRONMENT ${ATLAS_TEST_ENVIRONMENT}
+)
+
+ecbuild_add_test( TARGET atlas_test_interpolation_binning
+  SOURCES  test_interpolation_binning.cc
+  LIBS     atlas
+  MPI      6
```
wdeconinck, line 134: Add following as well please:
```
CONDITION eckit_HAVE_MPI AND MPI_SLOTS GREATER_EQUAL 6
```

src/atlas/interpolation/method/binning/Binning.cc
```diff
@@ -130,7 +137,8 @@ std::vector<double> Binning::getAreaWeights(const FunctionSpace& fspace) const {
   // diagonal of 'area weights matrix', W
   std::vector<double> ds_aweights;
 
-  if (fspace.type().compare(functionspace::detail::NodeColumns::static_type()) == 0) {
+  if (grid_type_ == "CS-LFR") {
```
wdeconinck, line 140: Possibly this logic can instead be, without configuration of grid_type:
```c++
bool is_cubed_sphere {false};
if (auto nodes_fs = functionspace::NodeColumns(fspace)) {
  if (CubedSphereGrid(nodes_fs.mesh().grid()) {
    is_cubed_sphere = true;
  }
}
if (is_cubed_sphere) { 
  ...
```
Then please remove grid_type_ everywhere.

wdeconinck (changes requested):
Thank you and sorry for the delayed review. Please see my in-code comments.

src/atlas/interpolation/method/binning/Binning.cc
```diff
@@ -35,7 +35,9 @@ MethodBuilder<Binning> __builder("binning");
 Binning::Binning(const Config& config) : Method(config) {
   const auto* ptr_config = dynamic_cast<const eckit::LocalConfiguration*>(&config);
   interpAncillaryScheme_ = ptr_config->getSubConfiguration("ancillary_scheme");
```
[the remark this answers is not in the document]
mo-lormi: @wdeconinck   unfortunately, this setup cannot be implemented.
To implement this constructor, I followed the same approach used for the constructor `SphericalVector()` implemented by @odlomax [here](https://github.com/ecmwf/atlas/blob/develop/src/atlas/interpolation/method/sphericalvector/SphericalVector.cc#L39).
<br>

src/atlas/interpolation/method/binning/Binning.cc
```diff
@@ -35,7 +35,9 @@ MethodBuilder<Binning> __builder("binning");
...
+  grid_type_ = ptr_config->getString("grid_type", "ATLAS");
```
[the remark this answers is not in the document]
mo-lormi: ... done

src/atlas/interpolation/method/binning/Binning.cc
mo-lormi: Good point ... thanks.
... done.

src/atlas/interpolation/method/binning/Binning.cc
```diff
@@ -0,0 +1,186 @@
...
+  target_ = target;
+
+  if (target_.size() == 0) {
+    return;
+  }
+
+  // enabling or disabling the halo exchange
+  this->allow_halo_exchange_ = halo_exchange_;
+  // enabling or disabling the adjoint operation
+  this->adjoint_ = adjoint_;
```
wdeconinck, line 68: This sets `adjoint_` to `adjoint_` (no-op). Probably not intentional?

src/atlas/interpolation/method/binning/Binning.cc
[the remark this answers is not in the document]
wdeconinck: Thanks I was confused and thought `Config` was `atlas::util::Config` but it is a typedef for `eckit::Parametrisation`.
For consistency with SphericalVector() I modified these lines in commit [d766d88](https://github.com/ecmwf/atlas/pull/191/commits/d766d889c7d2e8423eb400d67257623d7d9438c0) in this branch.

src/atlas/interpolation/method/binning/Binning.cc
mo-lormi: I implemented the adjoint operation as _optional_ -- _i.e._ if the operation is needed by the client, the client can set the flag to `true` and the setup of the adjoint operation will be carried out.
Do you see potential risks in this approach? ...
<br>

wdeconinck: Thanks I understand now. The problem is that the base class's `bool Method::adjoint_` which you were trying to set is private, and instead you were setting an overriding `bool Binning::adjoint_`.  I now made `bool Method::adjoint_`  protected (not private), and removed the overriding values in [b09c2a5](https://github.com/ecmwf/atlas/pull/191/commits/b09c2a5971765e4b8c2a1947dc4ab7245c87c3fc) so that the intended values are now modified.
I wonder why this bug was  failure did not occur before. Possibly some test was missing?
Also, @odlomax , I noticed a similar `bool SphericalVector::adjoint_` value in "SphericalVector" which may or may not be intending to overwrite the former private `Method::adjoint_`. Could you please verify that all is as expected in this regard?

odlomax: @wdeconinck,
Yep, that why I did it. It'll be safe to change it to the base class version once it's accessible.

wdeconinck: @odlomax --> [d31023e](https://github.com/ecmwf/atlas/pull/191/commits/d31023ed05151b46713f38cb7915237fb0ef8372)

src/atlas/grid/detail/grid/CubedSphere.h
```diff
@@ -241,9 +243,12 @@ class CubedSphere : public Grid {
     virtual std::string name() const override;
     virtual std::string type() const override;
 
-    // Return number of faces on cube
+    // Return N_, where (N_ * N_) is the number of cells on a tile
     inline idx_t N() const { return N_; }
 
+    // Return the areas of the faces/cells
+    Field gridCellArea(const FunctionSpace&) const;
```
wdeconinck, line 250: In order to encapsulate as much as possible the grid from objects like Field, FunctionSpace, Mesh,
I would avoid this function.

Possibly you could create a different function
```
double gridCellArea(idx_t index);
```
The grid itself should have all possible information needed to compute just a single cell value.

Then where needed fill a container such as `std::vector` or `atlas::Field` or ...
But in this case here, I think you could just compute the cellArea within the loop where you compute the weights, avoiding the entire allocation of a container.

mo-lormi: @wdeconinck  ... good point. However, there is reason why the function `CubedSphere::gridCellArea()` has been designed and implemented in this way.

So far, at the MO, we have identified at least two workflow steps where this function is needed: one is the evaluation of the _area weights matrix_ within the interpolation/binning procedure (this PR); the other one is the evaluation of the _Jc term_, a penalty term associated with the gravity wave constraints for the cost function.
Thus, we think that having a function signature for `CubedSphere::gridCellArea()` based on the `FunctionSpace` data type make it general enough to be used in different contexts.
<br>

wdeconinck: I managed to boil this down to this standalone function, independent of functionspace
```c++
// Area around nodes of cubed sphere mesh
Field CubedSphereNodalArea(const Mesh& mesh) {

  constexpr auto deg2rad = M_PI / 180.;
  const auto& proj = mesh.projection();
  auto lonlat = array::make_view<double, 2>(mesh.nodes().lonlat());
  auto gcell_area_field = Field("grid_cell_areas", make_datatype<double>(), array::make_shape(mesh.nodes().size()));  
  auto gcell_area_fview = array::make_view<double, 1>(gcell_area_field);

  double gcell_area_cs = [&] {
      /******* CUBED_SPHERE SPECIFIC *********/
      // (grid_res * grid_res) = no. of cells on a tile
      ATLAS_ASSERT(CubedSphereGrid(mesh.grid()));
      auto grid_res = CubedSphereGrid(mesh.grid()).N();
      // area of a grid cell (cubed-sphere coord. system)
      return M_PI/(2*grid_res) * M_PI/(2*grid_res);
   }();

  for (size_t i = 0; i < gcell_area_fview.size(); i++) {
    PointLonLat loc = PointLonLat(lonlat(i, atlas::LON), lonlat(i, atlas::LAT));
    double cos_lat = std::cos(deg2rad * loc.lat());
    double grid_jac_det = 1/proj.jacobian(loc).determinant();
    // area of a grid cell (geographic coord. system)
    gcell_area_fview(i) = grid_jac_det * gcell_area_cs * cos_lat ;   /****** USE gcell_area_cs here ******/
  }

  return gcell_area_field;
}
```
Possibly with more thought this could be made even more independent of CubedSphere.


What happens to the Binning procedure if the combination does not happen to be "grid == CubedSphere && functionspace == NodeColumns" ?

mo-lormi: @wdeconinck ... thanks for this.
Where do you think  this function should be stored? In which file, class,...?

wdeconinck: It could be a mesh-action, similar to https://github.com/ecmwf/atlas/blob/develop/src/atlas/mesh/actions/BuildCellCentres.h 
So putting it parallel to this file, and following its design is a good idea.
You can then "register" this field within the mesh.nodes() and could be reused later.

What happens to the Binning procedure if the combination does not happen to be "grid == CubedSphere && functionspace == NodeColumns" ?

mo-lormi:
<br>

@wdeconinck  thanks again for the comments and suggestions.
I re-implemented the procedure to evaluate the areas of the grid cells based on your proposal.
Thanks also to @odlomax  for helping to identify the cause of segmentation fault failure.
<br>

src/atlas/grid/detail/grid/CubedSphere.h
mo-lormi: > functionspace == NodeColumns

The _binning_ procedure works also for "functionspace == StructuredColumns" (grid type: "O").
I implemented a test to check this scenario, see [here](https://github.com/JCSDA-internal/atlas/blob/feature/interpolation_binning_01/src/tests/interpolation/test_interpolation_binning.cc#L155).

src/atlas/mesh/actions/GetCubedSphereNodalArea.cc
```diff
@@ -0,0 +1,56 @@
...
+#include "atlas/util/Point.h"
+
+
+namespace atlas {
+namespace mesh {
+namespace actions {
+
+
+Field& GetCubedSphereNodalArea::operator()(Mesh& mesh) {
+
```
wdeconinck, line 22: What's missing here is a check  if "grid_cell_areas" field exists in the mesh.nodes(), and return it if it exists, otherwise continue computing and adding it to the mesh.nodes() as you did already.

wdeconinck (approved):
Thanks so much for this contribution and addressing all my suggestions. 🙏

mo-lormi:
> Thanks so much for this contribution and addressing all my suggestions. 🙏

@wdeconinck thanks again.

wdeconinck:
This will be released in 0.38.0 some time this week.

Issue #192, opened 2024-05-03
Compile failure with NAG


DJDavies2:
### What happened?

I got this build failure with nagfor:

"/home/h01/frwd/cylc-run/mo-bundle-415/share/mo-bundle/atlas/src/atlas_f/field/atlas_Field_module.fypp", line 173: error: missing ".

repeated many times.

### What are the steps to reproduce the bug?

Compile with nagfor

### Version

develop

### Platform (OS and architecture)

Linux

### Organisation

Met Office

DJDavies2:
Fixed with https://github.com/ecmwf/fckit/commit/613d53ef5257db641ab103aefb93322f4456eb57.

PR #193, opened 2024-05-08
NAG fixes for fypp issue.

DJDavies2:
Resovles #192.

The problem is that when the pre-processor expands the macros the result contains lines of longer than 132 characters due to the length of the file paths.  This just adds some NAG specific code to circumvent the NAG compile issues described in #192.  It isn't a very good solution but I can't obviously see anything better.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
       - private-downstream-ci
       - downstream-ci-hpc
       - private-downstream-ci-hpc
-    if: always() && ${{ !github.event.pull_request.head.repo.fork && github.event.action != 'labeled' || github.event.label.name == 'approved-for-ci' }}
+    if: ${{ always() && !github.event.pull_request.head.repo.fork && github.event.action != 'labeled' || github.event.label.name == 'approved-for-ci' }}
     steps:
       - name: Trigger Teams notification
         uses: ecmwf-actions/notify-teams@v1
--- a/src/atlas/interpolation/method/Method.cc
+++ b/src/atlas/interpolation/method/Method.cc
@@ -301,7 +301,7 @@ void Method::setup(const FunctionSpace& source, const FunctionSpace& target) {
     ATLAS_TRACE("atlas::interpolation::method::Method::setup(FunctionSpace, FunctionSpace)");
     this->do_setup(source, target);
 
-    if (adjoint_) {
+    if (adjoint_ && target.size() > 0) {
         Matrix tmp(*matrix_);
 
         // if interpolation is matrix free then matrix->nonZeros() will be zero.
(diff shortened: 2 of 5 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/atlas?pullRequest=193) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/atlas?pullRequest=193) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/atlas?pullRequest=193) it.</sub>

wdeconinck:
I was just reading this here, which may solve this problem and be also a more desired output of handling `_FILE_` anyway
https://fypp.readthedocs.io/en/stable/fypp.html#rendering-file-names-as-relative-paths
I can have a look on adding this flag `--file-var-root` to fypp generation stage.

wdeconinck:
It doesn't seem to be so straightforward to apply a general approach with '--file-var-root'.
However I verified the generated file `<atlas-build-dir>/src/atlas_f/field/atlas_Field_module.F90`
and I don't get the line-length going over 132, and fypp has split the line in 2, e.g.:
```
  call fckit_exception%abort( "Rank mismatch", "/Users/willem/work/atlas-bundle/source/atlas/src/atlas_f/field/atlas_Field_module.f&
# 172 "/Users/willem/work/atlas-bundle/source/atlas/src/atlas_f/field/atlas_Field_module.fypp"
      &ypp", 172 )
  endif
  ```
Could you show me from the generated file as well an example of offending line?

DJDavies2:
> It doesn't seem to be so straightforward to apply a general approach with '--file-var-root'. However I verified the generated file `<atlas-build-dir>/src/atlas_f/field/atlas_Field_module.F90` and I don't get the line-length going over 132, and fypp has split the line in 2, e.g.:
> 
> ```fortran
>   call fckit_exception%abort( "Rank mismatch", "/Users/willem/work/atlas-bundle/source/atlas/src/atlas_f/field/atlas_Field_module.f&
> # 172 "/Users/willem/work/atlas-bundle/source/atlas/src/atlas_f/field/atlas_Field_module.fypp"
>       &ypp", 172 )
> ```
> 
> Could you show me from the generated file as well an example of offending line?

This is basically what I am seeing as well.  I think my description of the issue in the first comment isn't quite correct, the line is being chopped at 132 characters but the problem is the insertion of the # 172 ... line in the middle of the string line break.  It looks like NAG doesn't handle this syntax (it isn't standard).

wdeconinck:
Thanks for that clarification @DJDavies2 .
Perhaps you can try to add `FYPP_ARGS --line-numbering-mode=nocontlines` within this statement ? 
-> https://github.com/ecmwf/atlas/blob/develop/src/atlas_f/CMakeLists.txt#L279
And verify the continuation line "line markers" are gone?
If that is the case I can add this argument within the fypp macro within fckit for NAG compiler so that it would just work in any fypp file, not just Atlas.

DJDavies2:
FYPP_ARGS --line-numbering-mode=nocontlines does seem to work for the atlas_f subdirectory.  However there is also the file src/tests/runtime/fctest_trace.fypp and it isn't clear where I should make the change, can you point me to what I need to do?

wdeconinck:
Thanks for confirming it works. No need to change anything anymore.
I will make the change in the fypp cmake macro exported by fckit. That should then automatically fix atlas and other packages using fypp with NAG compiler.

wdeconinck:
That commit message in https://github.com/ecmwf/fckit/commit/613d53ef5257db641ab103aefb93322f4456eb57 auto-closed this PR. Interesting.... Please verify this works for you :)

DJDavies2:
Yes, that changed worked for me, thanks.

wdeconinck:
This is now released in fckit 0.13.0

PR #201, opened 2024-05-21
Add support for MultiField on GPU

sbrdar:
* Add FieldSet functions for device memory management via Field
* Add support for the copy-on-device of discontiguous host memory
* Add support for MultiField on the device


wdeconinck:
This PR will be split up in different ones.
- first up is #229
- then add GPU support via pluto
- then update remaining features and resolve conflicts.

PR #203, opened 2024-06-05
Remove float in Triag2D intersection algorithm

fmahebert:
Most floating-point numbers in this context are `double`, but there's a surprise `float` in the `Triag2D::intersects` algorithm. This lowers the precision of the `double` u,v coordinates.

I tried writing my own interpolation algorithm from the intersection outputs, and found I needed looser tolerances everywhere as a result of this `float` when using LonLat/2D triangles as compared to the Spherical/3D triangles.

Assuming this to be an oversight, I'm making this PR... let me know if the `float` is intentional.


wdeconinck:
Hi @fmahebert indeed that's bizarre, and should be double. Thanks for spotting!

Issue #205, opened 2024-06-06
Some interpolators fail with zero-sized target functionspace.


odlomax:
Sometimes, when performing an interpolation, situations arise where one or more partitions have zero-sized target functionspaces.

As reported in [jcsda oops #2664](https://github.com/JCSDA-internal/oops/issues/2664), an error occurs when the target functionspace of the Structured2D interpolation method has zero size. This triggers an error where the `eckit::SparseMatrix` fails to construct with zero non-zero elements. If we avoid constructing the interpolation object, MPI deadlocks can occur.

PR #206, opened 2024-06-06
Bugfix/allow zero-sized interpolation target functionspace

odlomax:
This PR addresses Atlas issue #205, and is needed for [jcsda/oops issue #2264](https://github.com/JCSDA-internal/oops/issues/2664).

The SphericalVector and Structured2D interpolation methods can now be called with empty target functionspaces. Also tested are the finite-element and cubedsphere interpolation methods.


wdeconinck (approved):
That looks good to me! I think there were already some guards like this in place, but probably only added in the methods where the issues were encountered and tested.

fmahebert:
@odlomax Thanks for this!

PR #208, opened 2024-06-07
Made sure cubed-sphere interpolation method always sets metadata.

odlomax:
Follow-on from #206.

Added a slight oversight in the last PR which, when combined with jcsda/oops [#2670](https://github.com/JCSDA-internal/oops/pull/2670), breaks the lfric-jedi model interface.

The metadata itself is a bit hacky and both will be redundant soon. But for now, it's needed to keep things running.

### DIFF
--- a/src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.cc
+++ b/src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.cc
@@ -33,6 +33,13 @@ void CubedSphereBilinear::do_setup(const FunctionSpace& source, const FunctionSp
     ATLAS_ASSERT(ncSource);
     ATLAS_ASSERT(target_);
 
+    // Add tile index metadata to target.
+    target_->metadata().set("tile index", std::vector<idx_t>{});
+
+    // Enable or disable halo exchange.
+    this->allow_halo_exchange_ = halo_exchange_;
+
+
     // return early if no output points on this partition reserve is called on
     // the triplets but also during the sparseMatrix constructor. This won't
     // work for empty matrices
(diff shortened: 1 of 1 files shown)


src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.cc
```diff
@@ -33,6 +33,13 @@ void CubedSphereBilinear::do_setup(const FunctionSpace& source, const FunctionSp
     ATLAS_ASSERT(ncSource);
     ATLAS_ASSERT(target_);
 
+    // Add tile index metadata to target.
+    target_->metadata().set("tile index", std::vector<idx_t>{});
```
wdeconinck, line 37: It doesn't look right that we're modifying the target function space metadata here, especially since it should be `const`.

wdeconinck: This should be changed downstream:
```
std::vector<idx_t> tile_index;
target_fs.metadata().get("tile index", tile_index);
```

odlomax: Fair. I was being lazy because it's already being set on line 109 of this file. I can remove it. Two wrongs don't make a right.

I plan on removing this and the `allow_halo_exchange_` boolean once the oops PRs are merged. It's terribly hacky.

wdeconinck: So you agree to modify this PR to correct the wrongs, and leave this hanging until oops PRs are fixed? I'm not entirely clear on your intention.

odlomax: Sorry, I'm communicating poorly.

This PR:
I'll revert the metadata change on line 37.

Next one (once oops PRs are settled):
I'll remove the line 109 metadata and the halo exchange control. This will involve a significant amount of downstream changes, so It'll take a bit of time.

odlomax:
This one has a green light from me. I'll create an issue with the other tidy-up items I mentioned.

wdeconinck (approved):

odlomax:
Thank you kindly, @wdeconinck.

Would it be possible to issue a tag once this and #191 have gone in?

wdeconinck:
This will be released in 0.38.0 some time this week.

PR #209, opened 2024-06-10
Add Fortran interface for node-to-edge connectivity building

benjaminmenetrier:
The Fortran interface for node-to-edge connectivity building was half-implemented, this PR makes it fully available.


wdeconinck (approved):
Thanks @benjaminmenetrier . Very uncontroversial.

Issue #194, opened 2024-05-15
Risk of surprising behavior when adding multiple Fields of same name into FieldSet


fmahebert:
### What happened?

It's possible to add multiple Fields sharing the same name into one FieldSet. This is already surprising, because it's not immediately clear that `fieldset[name]` would be well defined.

Looking through the code, I believe it _is_ well-defined in the current implementation: the `FieldSetImpl::add` method updates the FieldSet's internal mapping `index_` from names to indices to add the name of the new Field being added; but when that name is a duplicate, the result is to update the map element to point to a new index instead of _adding_ a new map element. See here,
https://github.com/ecmwf/atlas/blob/55e20456d5f754e9d6e1180d8db5fa866010bf3b/src/atlas/field/FieldSet.cc#L59

The result is that the most-recently-inserted Field with some name will be the one returned by `fieldset[name]`. See reproducer below.

This opens the door for various inconsistencies. For example,
```
for (const auto & name : fieldset.field_names()) {
  auto & field = fieldset[name];
}
```
will not actually loop over all Fields — if there's a N-fold duplicate name it will hit the most-recently-inserted Field N times.

Is allowing duplicate Fields of the same name intentional? I believe the code would be safer if this were simply forbidden.

### What are the steps to reproduce the bug?

Here's code that shows that the most-recently-inserted Field is the one returned
```
#include <iostream>
#include "atlas/field.h"
#include "atlas/functionspace.h"
#include "atlas/grid.h"
#include "atlas/util/Config.h"
#include "eckit/runtime/Main.h"

int main(int argc, char **argv) {
  using namespace atlas;
  eckit::Main::initialise(argc, argv);

  const Grid grid("N16");
  const functionspace::StructuredColumns fs(grid);

  Field f1 = fs.createField<double>(option::name("name") | option::levels(1));
  Field f2 = fs.createField<double>(option::name("name") | option::levels(2));
  Field f3 = fs.createField<double>(option::name("name") | option::levels(3));
  FieldSet fieldset;
  fieldset.add(f1);
  fieldset.add(f2);
  fieldset.add(f3);
  for (const auto & name : fieldset.field_names()) {
    std::cout << "field: " << name << " has levels: " << fieldset[name].levels() << std::endl;
  }

  return 0;
}
```

### Version

0.35

### Platform (OS and architecture)

macOS

MarekWlasak:
[...]

PR #210, opened 2024-06-12
Avoid silent errors accessing Fieldset fields by ambiguous names

wdeconinck:
This should address #194 

An exception is thrown when accessing a field by name in the fieldset, when the name is ambiguous.
A unit-test is created that demonstrates the exception.

### DIFF
--- a/src/atlas/field/FieldSet.cc
+++ b/src/atlas/field/FieldSet.cc
@@ -22,11 +22,11 @@ namespace field {
 //------------------------------------------------------------------------------------------------------
 
 void FieldSetImpl::FieldObserver::onFieldRename(FieldImpl& field) {
-    std::string name = field.name();
-    for (auto& kv: fieldset_.index_) {
-        const auto old_name = kv.first;
-        const auto idx      = kv.second;
-        if (&field == fieldset_.fields_[idx].get()) {
+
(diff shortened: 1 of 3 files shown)


DJDavies2:
I seem to be getting a number of failures in downstream codes with this like so:

Exception: FieldSet: cannot get field with ambiguous name 
height_levels". 2 fields are registered with same name. Access field by index or iterator instead.  (/home/h01/david.davies/cylc-run/atlas-210/share/mo-bundle/atlas/src/atlas/field/FieldSet.cc +120 field)
FieldSet: cannot get field with ambiguous name 
height_levels". 2 fields are registered with same name. Access field by index or iterator instead. caught in  (/home/h01/david.davies/cylc-run/atlas-210/share/mo-bundle/oops/src/oops/runs/Run.cc +197 execute)
Exception: oops::ErrorCovarianceToolbox<quench> terminating...
Exception stack: 
FieldSet: cannot get field with ambiguous name 
height_levels". 2 fields are registered with same name. Access field by index or iterator instead.

wdeconinck:
Thanks @DJDavies2 , this are exactly the silent errors mentioned in #194 (notifying @fmahebert @svahl991)

wdeconinck:
@svahl991 I updated the testing :)

fmahebert:
@wdeconinck Thanks for this!

PR #215, opened 2024-07-26
Add regional interpolation

benjaminmenetrier:
This PR adds a new interpolation for regional structured source grids, available with the factory tag "regional-linear-2d". It uses a generic sparse matrix element class `InterpElement` added in `interpolation/element`, that might also be used for vertical interpolation in the future.

An associated test has been added, which defines field values on a coarse source grid, interpolates on a refined target grid, and compute the target field hash on each MPI task (2 tasks requested for this test).

The PR also includes a bugfix in `functionspace/detail/StructuredColumns_setup.cc`: it seems that the field `xy` is actually corresponding to the `lonlat` values for all structured grids. However, this was not true in `develop` for regional grids. I have implemented a minimal bugfix where the correct values of `lonlat` are specified in the `xy` field for regional grids too.


benjaminmenetrier:
It seems the hash test is failing depending on the compiler, I'll do it differently.

benjaminmenetrier:
The new test is working.

benjaminmenetrier:
I have extended the regional interpolation test : accuracy and adjoint tests, for both 1D and 2D fields.

src/atlas/interpolation/method/structured/RegionalLinear2D.cc
```diff
@@ -0,0 +1,557 @@
...
+        // Bilinear interpolation factor
+        const double alphaX = 1.0-(sourceGrid.x(indexI)+sourceDx-targetX)/sourceDx;
+        const double alphaY = reversedY ? (sourceGrid.y(indexJ)-targetY)/sourceDy
+          : 1.0-(sourceGrid.y(indexJ)+sourceDy-targetY)/sourceDy;
+
+        // Points to find
+        std::vector<bool> toFind = {true, !colocatedX, !colocatedY, !colocatedX && !colocatedY};
+        std::vector<size_t> valueToFind = {indexI*sourceNy+indexJ, (indexI+1)*sourceNy+indexJ,
+          indexI*sourceNy+(indexJ+1), (indexI+1)*sourceNy+(indexJ+1)};
+        std::vector<int> foundIndex(4, -1);
```
wdeconinck, line 317: Since these are always size 4, it will be much better to use ```std::array<type,4>``` instead of `std::vector<type>`

src/atlas/interpolation/method/structured/RegionalLinear2D.cc
```diff
@@ -0,0 +1,557 @@
...
+              (std::abs(targetY-sourceGrid.y(indexJ+1)) < std::abs(targetY-sourceGrid.y(indexJ)))) {
+              indexJ += 1;
+            }
+          } else {
+            if (std::abs(targetY-sourceGrid.y(0)) < std::abs(targetY-sourceGrid.y(sourceNy-1))) {
+              indexJ = 0;
+            } else {
+              indexJ = sourceNy-1;
+            }
+            std::cout << "WARNING: point outside of the domain" << std::endl;
```
wdeconinck, line 300: Please use "Log::info()" or "Log::error()"

src/atlas/interpolation/method/structured/RegionalLinear2D.h
```diff
@@ -0,0 +1,76 @@
...
+  std::vector<int> mpiTask_;
+  size_t targetSize_;
+  size_t sourceSendSize_;
+  size_t targetRecvSize_;
+  std::vector<int> sourceSendCounts_;
+  std::vector<int> sourceSendDispls_;
+  std::vector<int> targetRecvCounts_;
+  std::vector<int> targetRecvDispls_;
+  std::vector<size_t> sourceSendMapping_;
+  std::vector<atlas::interpolation::element::InterpElement> horInterp_;
```
wdeconinck, line 70: `std::vector<atlas::interpolation::element::InterpElement>` in current implementation of `InterpElement` is essentially the same as
`std::vector<std::vector<std::pair<size_t,double>>>`

The issue is `std::vector<std::vector< ... >>` because that infers multiple allocations on the heap within tight loop and slow memory access.
I'm thinking to avoid the `InterpElement` class altogether (and remove the file etc.)

What would be better is multiple arrays:
```c++
std::vector< std::array<size_t,4> > stencil_;
std::vector< std::array<double,4> > weights_;
std::vector< size_t > stencil_size_;
```
and then resize all these arrays to `targetSize_` before looping over target points, and fill the values at the correct place in these arrays, without using the temporary `operations` variable. This will also allow this loop to become OpenMP multi-threadable.

These 3 vectors are now each one large memory allocation because the memory of `std::array<type,4>` is known at compile time.
We need the `stencil_size_` array to avoid going into possibly over-allocated entries in case of linear interpolation or point value, which will almost never be the case.

src/atlas/functionspace/detail/StructuredColumns_setup.cc
```diff
@@ -588,13 +588,20 @@ void StructuredColumns::setup(const grid::Distribution& distribution, const ecki
...
             const GridPoint& gp = gridpoints[n];
-            if (gp.j >= 0 && gp.j < grid_->ny()) {
-                xy(gp.r, XX) = grid_->x(gp.i, gp.j);
-                xy(gp.r, YY) = grid_->y(gp.j);
-            }
-            else {
-                xy(gp.r, XX) = compute_x(gp.i, gp.j);
-                xy(gp.r, YY) = compute_y(gp.j);
+            if (regional) {
+              std::vector<double> lonlatVec(2);
```
wdeconinck, line 592:
```suggestion
              std::array<double,2> lonlatVec;
```
This avoids heap allocation within the loop over grid points, and should give very nice speedup.

wdeconinck (changes requested):
This is a very nice addition @benjaminmenetrier and will be very useful! 
This is just a first review without having played around with this feature yet, and sorry that it has taken this long.
Just looking through the code I could find some performance / memory issues that need to be addressed. Please have a look.

src/atlas/interpolation/method/structured/RegionalLinear2D.cc
benjaminmenetrier: Done.

src/atlas/interpolation/method/structured/RegionalLinear2D.cc
benjaminmenetrier: Done.

src/atlas/interpolation/method/structured/RegionalLinear2D.h
benjaminmenetrier: Great suggestion. For regional DA, we often interpolate from a grid to another one with exactly the same rectangular domain, and just a different cell size, so `stencil_size_` might be lower than 4 quite often.

src/atlas/functionspace/detail/StructuredColumns_setup.cc
benjaminmenetrier: Done.

benjaminmenetrier:
Thank you very much @wdeconinck for your very relevant comments. I realize that I am not well aware of memory usage in C++ and should be more careful about it.

wdeconinck (approved):
Thank you for this nice development, and thank you for addressing the suggestions.

PR #219, opened 2024-08-21
Add HIC abstraction layer for HIP and CUDA

wdeconinck:
Currently this is implemented as a stand-alone CMake package within the atlas repository.

A dummy API is available when HIP or CUDA backend is not available, to avoid `#ifdef` in downstream code.

Atlas has been adapted to replace all use of CUDA API with the new hic API.
Even though hic is tested with HIP backend, the use of Atlas with hic has not been tested with the HIP backend. A future PR may address small fixes needed.


src/atlas/array/native/NativeDataStore.h
```diff
@@ -213,6 +213,7 @@ class DataStore : public ArrayDataStore {
 #if ATLAS_HAVE_ACC
         if (not acc_mapped_) {
             ATLAS_ASSERT(deviceAllocated(),"Could not accMap as device data is not allocated");
+            atlas_acc_unmap_data(host_data_);
```
wdeconinck, line 216: Need to remove above line.

PR #222, opened 2024-09-12
Bug fixes to vector interpolation with StructuredColumns and spherical vector interpolation

MarekWlasak:
This is a set of minor bug-fixes

The most notable (and contentious) is that for the vector interpolation it is currently assuming that the `levels` have been given to StructuredColumns at construction. For most functionspaces this is not the case, so I have added a small change that will allow the changing of the sign in the 3D case, when `levels` is not set in the functionspace but instead in the field.

I have added a number of additional tests to `test_interpolation_spherical_vector.cc` including the classic Gauss state and cubic interpolation.

### DIFF
--- a/src/atlas/functionspace/detail/StructuredColumns.cc
+++ b/src/atlas/functionspace/detail/StructuredColumns.cc
@@ -540,7 +540,7 @@ const util::PartitionPolygon& StructuredColumns::polygon(idx_t halo) const {
     }
     if (not polygons_[halo]) {
         if (halo > this->halo()) {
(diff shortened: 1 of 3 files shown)


MarekWlasak:
@wdeconinck - looks like a private-downstream-ci-hpc  failed.  Do I need to increase the tolerance on any of the new tests?

wdeconinck:
> @wdeconinck - looks like a private-downstream-ci-hpc failed. Do I need to increase the tolerance on any of the new tests?

Thanks @MarekWlasak it looks like it's not related to Atlas but one specific cmake configuration failing, so nothing to do on your part.

@dtip or @figi44 could you have a look why this occurs?

PR #223, opened 2024-09-24
Fixed ordering of `fixup_halos` and `halo_exhange.execute_adjoint` in `StructuredColumns.cc`

odlomax:
Whilst integrating `util::pack_vector_fields` into the `SphericalVector` interpolation method, I uncovered a bug in `StructuredColumns`. The methods `fixup_halos` and `halo_exchange.execute_adjoint` were called in the wrong order during a halo exchange adjoint. This PR fixes the bug.

I'd initially missed this by adding (and then forgetting) a workaround in the `SphericalVector` test.

@MarekWlasak this will likely be of interest to you.


odlomax:
@wdeconinck Am I correct in thinking the test failure isn't related to these commits? Let me know if there's anything I can do.

wdeconinck:
I would like to see what @MarekWlasak has to say about this, otherwise it is good to merge for me.

odlomax:
@MarekWlasak, please could you bless or condemn this PR!

PR #220, opened 2024-08-22
Added function to create `std::variant` for multiple array views.

odlomax:
During PR #218, I found myself writing two nested switch statements in order to create an `ArrayView` from an `Array`. I've found myself writing this same pattern multiple times throughout Atlas, and indeed other external applications which use atlas.

This PR adds provides an alternative to the nested switch statement. The method `util::make_array_view_variant<Values<types...>, Ranks<ints...>>(Array& array)` creates an `std::variant` of `ArrayView` types, and then assigns the view that matches `array` (so long as the value-type and rank can be found within the `Values` and `Ranks` arguments respectively. The `ArrayView` can then be accessed rather succinctly using `std::visit`.

PR #224, opened 2024-09-24
Integrate `pack_vector_fields` into `SphericalVector` Interpolation method.

odlomax:
Hi @wdeconinck 

This is part of issue #217. Here we've integrated the helper method from PR #218 into the `SphericalVector` interpolation method. Another `CASE` has been added to the interpolation test which demonstrates vector interpolation with components stored in individual fields.

Note: PR #223 needs to be merged before this.

@MarekWlasak, this is how I envisaged the use of this interpolation method within JEDI where we exclusively store vector field components in separate fields. 

The final PR for issue #217 will integrate PR #220 into the helper and interpolation methods used here.


odlomax:
I've rebased this one off of the head of `develop`. It's ready for review whenever you are. :-)

wdeconinck (approved):
Very nice and elegant. Slowly making my way through your PRs.

PR #226, opened 2024-09-25
Refactor `util::pack_vector_fields` to use `array::ArrayViewVariant`

odlomax:
This PR refactors `util::pack_vector_fields` to use `std::variant<ArrayViews...>` with `std::visit`, as opposed to nested `switch` statements to resolve the possible `Value` types and `Rank` values for different arrays.

I also found it beneficial to refactor `array::make_view_variant` such that it produces either a variant of `const` or a variant of `non- const` ArrayView types.

This PR needs #220 and #224 to be merged first.

PR #227, opened 2024-09-25
Refactor `SphericalVector` interpolation method to use `array::ArrayViewVariant`.

odlomax:
This is the last of this batch of PRs that refactors the `SphericalVector` interpolation method. Here we've used `std::visit` on `ArrayViewVariant` to resolve the rank and type of the arrays instead of several verbose `if` statements and function overloads.

This PR depends on PRs #220, #223, #224 and #226. It closes issue #217.


wdeconinck (approved):
nice!

PR #230, opened 2024-10-09
Bugfix for Qhull

benjaminmenetrier:
Hi @wdeconinck, I think the `atlas-import.cmake.in` file should be updated since Qhull is used instead of CGAL for the TESSELATION option. Is it correct?


wdeconinck:
Hi @benjaminmenetrier you're right, good catch! It should normally only matters for static builds though.

benjaminmenetrier:
> Hi @benjaminmenetrier you're right, good catch! It should normally only matters for static builds though.

Yes, that's how I noticed it! Our Harmonie system has such a static build (not my choice).

PR #231, opened 2024-10-10
Tidied up headers in `ArrayViewVariant.h` and `ArrayViewVariant.cc`

odlomax:
This PR does some tidying that I missed in PR #220. Namely headers and the error message. I also found a missing `<sstream>` header in `runtime/Exception.h`.


wdeconinck:
Nice @odlomax , appreciate this! I will merge this immediately when tests have run through.

PR #232, opened 2024-10-10
Add Fortran access to device data through Field and FieldSet

sbrdar:
* add access to the device pointer in atlas_Field_module and atlas_FieldSet_module for the 'deviceptr' clause of OpenACC
* add sync_host_device;


wdeconinck (approved):
Reviewed offline and changes have been pushed.
Thanks for this nice work allowing us to now use `!$acc data deviceptr` clause in Fortran!

PR #235, opened 2024-11-07
Move example-fortran inside atlas_HAVE_DOCS

benjaminmenetrier:
Hi @wdeconinck, would it be possible to move the `example-fortran` code compilation within the `atlas_HAVE_DOCS` key? This would be helpful for me because I need to compile the Fortran library, but I don't want to link the `example-fortran` binaries. Thanks!


wdeconinck (approved):
Right that was an oversight! Thanks for this.

PR #240, opened 2024-11-19
Extend sparse linear algebra interface with multiply-add

l90lpa:
This PR extends Atlas's sparse linear algebra interface with multiply-add functionality.

### DIFF
--- a/src/atlas/interpolation/method/Method.cc
+++ b/src/atlas/interpolation/method/Method.cc
@@ -168,65 +168,30 @@ void Method::interpolate_field_rank3(const Field& src, Field& tgt, const Matrix&
 
 template <typename Value>
 void Method::adjoint_interpolate_field_rank1(Field& src, const Field& tgt, const Matrix& W) const {
-    array::ArrayT<Value> tmp(src.shape());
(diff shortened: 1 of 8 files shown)


l90lpa:
Removed dependency on PR https://github.com/ecmwf/atlas/pull/237

l90lpa:
Hi @wdeconinck, this PR currently only exposes the simple $y=Ax+y$ interface. I'm wondering if you have an actual need for $y=\alpha Ax+\beta y$ and if not whether we could leave this as follow-on action?

wdeconinck:
Thanks @l90lpa for this PR. I will review it.

> Hi @wdeconinck, this PR currently only exposes the simple y = A x + y interface. I'm wondering if you have an actual need for y = α A x + β y and if not whether we could leave this as follow-on action?

Hi @l90lpa I think it's OK indeed to leave it as a follow-on action if ever needed, but as you say, unlikely.

src/atlas/interpolation/method/Method.cc
wdeconinck, line 1: Nice cleanup with new API

wdeconinck (approved):
I think it's really good and cleans up quite a nice chunk in the adjoint code path, and will improve performance! This could be merged as is a.f.a.i.c

PR #241, opened 2024-11-20
Add GPU offload capable sparse matrix

l90lpa:
This PR adds a sparse matrix class, with GPU memory management functionality similar to Array. The class uses eckit's SparseMatrix class internally for host-side memory storage and expose an interface similar to that of eckit's SparseMatrix class.

### DIFF
--- a/hic/tests/test_hicsparse.cc
+++ b/hic/tests/test_hicsparse.cc
@@ -283,6 +283,7 @@ int test_hicsparseSpMM() {
     }
 
     // Clean up
+    HIC_CALL(hicFree(buffer));
     HIC_CALL(hicFree(dC));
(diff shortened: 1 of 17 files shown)


l90lpa:
Hi @wdeconinck thanks for taking a look and adding that extra test. I've opened https://github.com/ecmwf/atlas/pull/246 as the next follow-on PR.

wdeconinck:
Thanks @l90lpa this was an excellent push in the right direction. I have taken the liberty to redesign this in #247 but very much inspired by this PR by splitting the responsibility of creation/storage/view, into different classes or methods, making it more flexible and maintainable for future redesign and compatibility with various users.
I will close this PR now.

PR #246, opened 2024-11-26
Add a hicsparse backend to linalg sparse

l90lpa:
This PR adds a hicsparse backend to the sparse matrix multiplication routines of linalg sparse.

This PR depends on https://github.com/ecmwf/atlas/pull/247 and the branch of this PR is currently based on the branch of https://github.com/ecmwf/atlas/pull/247.

Notes:
- This PR previously depended on https://github.com/ecmwf/atlas/pull/241 but that dependency is nolonger necessary as https://github.com/ecmwf/atlas/pull/247 will superseded the former PR.

### DIFF
--- a/src/atlas/CMakeLists.txt
+++ b/src/atlas/CMakeLists.txt
@@ -712,6 +712,8 @@ linalg/sparse/SparseMatrixMultiply.h
 linalg/sparse/SparseMatrixMultiply.tcc
 linalg/sparse/SparseMatrixMultiply_EckitLinalg.h
 linalg/sparse/SparseMatrixMultiply_EckitLinalg.cc
+linalg/sparse/SparseMatrixMultiply_HicSparse.h
+linalg/sparse/SparseMatrixMultiply_HicSparse.cc
 linalg/sparse/SparseMatrixMultiply_OpenMP.h
 linalg/sparse/SparseMatrixMultiply_OpenMP.cc
 linalg/sparse/SparseMatrixStorage.cc
@@ -1007,6 +1009,7 @@ ecbuild_add_library( TARGET atlas
     eckit_option
     atlas_io
     hic
+    hicsparse
     $<${atlas_HAVE_EIGEN}:Eigen3::Eigen>
     $<${atlas_HAVE_OMP_CXX}:OpenMP::OpenMP_CXX>
     $<${atlas_HAVE_GRIDTOOLS_STORAGE}:GridTools::gridtools>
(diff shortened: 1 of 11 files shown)


l90lpa:
I've rebased this PR as its dependency has now changed from https://github.com/ecmwf/atlas/pull/241 to https://github.com/ecmwf/atlas/pull/247 (see comment https://github.com/ecmwf/atlas/pull/247#issuecomment-2520799855).

wdeconinck (changes requested):
This looks nearly perfect. Very nice work, thanks! 🙏
I rebased it on develop which now contains the dependent #247 .
What's missing still is support for having the SourceValue and TargetValue be of a different precision than the MatrixValue.
This is e.g. the case with the OpenMP backend. The matrix is typically double precision (eckit::linalg::Scalar), whereas the SourceValue and TargetValue come from the atlas Fields which can be single or double.
Check e.g. this hicsparse_matrix_multiply in https://github.com/ecmwf/atlas/blob/develop/src/tests/linalg/test_linalg_sparse_matrix_gpu.cc#L32 showing the different precisions.
Note that alpha and beta arguments need to be of a chosen "ComputeType". In above linked routine I chose ComputeType to match the MatrixValue type, but in principle that is another user-choice that could be made...

wdeconinck:
Possibly this hicsparse backend only works for fields which are contiguous in memory, e.g. not separate slices of a larger array with stride != 1.
We may need to abort if that is the case. It would be good to test.

l90lpa:
@wdeconinck thanks for taking a look at this PR! Apologies for the slow reply, I've been on leave.

- **Regarding mixed field types**:
  - I don't think the choice of types, like for `compute type`, is quite as flexible as you're suggesting. If you look in the cuSparse documentation for say `cusparseSpMM` it doesn't suggest that you can mix `float32` and `float64` types - which lines up with what I've seen when I experimented with it (however in the case of `cusparseSpMV` it looks like you can mix a `float32` sparse matrix with `float64` vectors but I decided to only support the common set to both `cusparseSpMV` and `cusparseSpMM`).
  - the link you sent me shows an example of how you might wrap the `cusparseSpMV` to admit `m`, `x` and `y` of different types (something that I started with before simplifying it due to precision mixing restrictions), but the only use of the wrapper looks to be inside `check_matrix` which only uses the wrapper in a uniform type way. You could perhaps modify `check_matrix` if you want to experiment with mixing different types - perhaps there are some combinations that are not documented.
- **Regarding the non-contiguous in memory fields**: the `cuspareSpMM` api does support non-contiguous in memory dense matrices and this is handled by using `leadingDimension` parameter of `cusparseCreateDnMat` when setting up a dense matrix descriptor (which I think I have already made use of). I think the primary use case being to pass in a submatrix of a larger matrix. Unless are you suggesting a case where neither axis of the matrix is contiguous in memory?

wdeconinck (approved):

PR #248, opened 2024-12-04
Offloading discontiguous host data

sbrdar:
* offloading discontiguous data via WrappedDataStorage
* Field can wrap discontiguous array and copy it to a contiguous memory on the device

### DIFF
--- a/src/atlas/array/native/NativeArray.cc
+++ b/src/atlas/array/native/NativeArray.cc
@@ -62,7 +62,7 @@ Array* Array::wrap(Value* data, const ArrayShape& shape) {
 template <typename Value>
 Array* Array::wrap(Value* data, const ArraySpec& spec) {
     size_t size = spec.size();
-    return new ArrayT<Value>(new native::WrappedDataStore<Value>(data, size), spec);
+    return new ArrayT<Value>(new native::WrappedDataStore<Value>(data, spec, size), spec);
 }
 
 Array::~Array() = default;
--- a/src/atlas/array/native/NativeArray.cc
+++ b/src/atlas/array/native/NativeArray.cc
@@ -62,7 +62,7 @@ Array* Array::wrap(Value* data, const ArrayShape& shape) {
 template <typename Value>
 Array* Array::wrap(Value* data, const ArraySpec& spec) {
     size_t size = spec.size();
-    return new ArrayT<Value>(new native::WrappedDataStore<Value>(data, spec, size), spec);
+    return new ArrayT<Value>(new native::WrappedDataStore<Value>(data, spec), spec);
 }
 
 Array::~Array() = default;
--- a/src/atlas/array/native/NativeDataStore.h
+++ b/src/atlas/array/native/NativeDataStore.h
@@ -241,7 +241,7 @@ private:
 template <typename Value>
 class WrappedDataStore : public ArrayDataStore {
 public:
-    WrappedDataStore(Value* host_data, const ArraySpec& spec, size_t size): host_data_(host_data), size_(size),
+    WrappedDataStore(Value* host_data, const ArraySpec& spec): host_data_(host_data), size_(spec.size()),
         device_memory_resource_(memory::device::traced_resource()),
         device_allocator_{device_memory_resource_.get()}
     {
@@ -427,7 +427,7 @@ private:
 
     mutable Value* device_data_;
 
-    bool contiguous_;
+    bool contiguous_{true};
     size_t memcpy_h2d_pitch_;
     size_t memcpy_d2h_pitch_;
     size_t memcpy_height_;
--- a/src/atlas/array/native/NativeDataStore.h
+++ b/src/atlas/array/native/NativeDataStore.h
@@ -241,9 +241,49 @@ private:
 template <typename Value>
 class WrappedDataStore : public ArrayDataStore {
 public:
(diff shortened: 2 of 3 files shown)


src/atlas/array/native/NativeDataStore.h
```diff
@@ -241,9 +241,49 @@ class DataStore : public ArrayDataStore {
 template <typename Value>
 class WrappedDataStore : public ArrayDataStore {
 public:
+    WrappedDataStore(Value* host_data, const ArraySpec& spec, size_t size): host_data_(host_data), size_(size),
```
wdeconinck, line 244: The `size` argument is redundant with `spec` and could lead to inconsistencies.
I propose to extract size from `spec` instead.

src/atlas/array/native/NativeDataStore.h
```diff
@@ -241,9 +241,49 @@ class DataStore : public ArrayDataStore {
...
+            else {
+                memcpy_h2d_pitch_ = shp_mult_rhs;
+                memcpy_d2h_pitch_ = spec.strides()[break_idx];
+                memcpy_width_ = shp_mult_rhs;
+                memcpy_height_ = shp_mult_lhs;
+            }
+        }
+    }
+
     WrappedDataStore(Value* host_data, size_t size): host_data_(host_data), size_(size),
```
wdeconinck, line 279: This constructor does not set a default value for `contiguous_`.
I don't think this constructor is used anymore as its use is changed for the one using `spec`, but regardless good to keep and fix.

src/tests/field/fctest_field_wrap_gpu.F90
```diff
@@ -90,39 +90,45 @@ module fcta_Field_wrap_device_fixture
...
 
-  !!$acc data present(fview)
-  !!$acc parallel loop
-  do i=1,4
-    do j=1,3
-      do k=1,2
-        do l=1,5
-          fview(i,j,l) = 1000*i+100*j+10*1+l
-        enddo
+  !$acc parallel loop deviceptr(fview)
```
wdeconinck, line 116: I'm afraid that this `deviceptr(fview)` with fview having the `POINTER` attribute  is non-standard OpenACC and will not compile with Cray compiler. We need to create a separate subroutine just for kernels with deviceptr  😞, where the deviceptr is passed as argument without `POINTER` attribute.

src/tests/field/fctest_field_wrap_gpu.F90
```diff
@@ -90,39 +90,45 @@ module fcta_Field_wrap_device_fixture
...
 
-  k=1
-  do i=1,4
-    do j=1,3
-      do l=1,5
-        FCTEST_CHECK_EQUAL(fview(i,j,l) , real(1000*i+100*j+10*k+l,c_double) )
+  do l = 1, Nl
+    do j = 1, Nj
+      do i = 1, Ni
+        FCTEST_CHECK_EQUAL(existing_data(i, j, k_idx, l) , real(-1000*i - 100*j - 10*k_idx - l, c_double) )
```
wdeconinck, line 131: It would be good to add the extra test that verifies existing_data is not overwritten for `k != k_idx`.

src/atlas/array/native/NativeArray.cc
```diff
@@ -62,7 +62,7 @@ Array* Array::wrap(Value* data, const ArrayShape& shape) {
 template <typename Value>
 Array* Array::wrap(Value* data, const ArraySpec& spec) {
     size_t size = spec.size();
-    return new ArrayT<Value>(new native::WrappedDataStore<Value>(data, size), spec);
+    return new ArrayT<Value>(new native::WrappedDataStore<Value>(data, spec, size), spec);
```
wdeconinck, line 65: So we can remove the extraction of `size` in this case, and only pass `spec` to WrappedDataStore

wdeconinck (changes requested):
Very good! I like how little code it was to add based on `pluto` implementation.

I made some very minor suggestions in code.

src/tests/field/fctest_field_wrap_gpu.F90
sbrdar: I have added the extra test in b98e3bca7

src/atlas/array/native/NativeDataStore.h
sbrdar: I agree and am extracting `size` from `spec` now. See 6811aa37f. Thanks!

src/atlas/array/native/NativeDataStore.h
sbrdar: Indeed, thank you. `contiguous_` is now defaulted to `true`, except if `spec` is used in the constructor and then it is `spec.contiguous()`. See 6811aa37f.

src/atlas/array/native/NativeArray.cc
sbrdar: Thanks for the comment. I have taken your suggestion directly in 6811aa37f.

src/tests/field/fctest_field_wrap_gpu.F90
sbrdar: This was missing. Thank you for pointing it out. I have added kernel subroutines to call ACC pragmas. See 5dc24afbb.

sbrdar:
@wdeconinck Thank you for the comments!

wdeconinck (approved):
Thank you for addressing all the changes! Good to go!

PR #257, opened 2025-02-06
Python wheel: initial setup

tmi:


iainrussell:
@wdeconinck, this is part of our effort to provide a large portion of our software stack as Python wheels. We have a new convention, where the binary libraries are stored on PyPi as <package>lib, e.g. eccodeslib, mirlib, fdblib, up to multio. We would add atlaslib to this, so that mirlib can be built with atlas support. It's still a bit experimental, but we hope that earthkit-regrid can soon start benefiting from it. I'm happy to merge this in so that we can start using the GitHub actions to deploy it, if you give us the thumbs-up!

tmi:
There's one particular problem here: https://pypi.org/project/atlaslib/ is already taken by someone, so we need a different name here. No preference on my side, so something like "ecatlaslib"? Ftr `atlas` is also taken since Dec 2024 :/ 

This  unfortunately means that `findlibs` invocation would have to be done via `findlibs.find("atlas", <the new name>)` instead of just `findlibs.find("atlas")`

wdeconinck:
Hi, the new name could be something similar to what we have used for readthedocs.
There it is "atlas-ecmwf.readthedocs.io"
So this could be "atlaslib-ecmwf" ?

wdeconinck:
Hi @tmi could you please move this directory into ".github/python_wrapper_wheel" or so?
It should not be confused with the existence of an atlas Python API at this point.

tmi:
@wdeconinck the convention we've started in all other ecmwf projects is to have the wrapper in `python_wrapper` and the python API (as in, proper interfaces callable from python, with python objects) in `python_interface` directory. See eg https://github.com/ecmwf/metkit/pull/55/files. That can of course be changed, ideally consistently across all the repositories -- `python_wrapper` is now in about 10... However I'm not sure why you would put it to `.github`, it does not have anything to do with github, does it? Or is it more about just hidding the directory somehow?

tmi:
@wdeconinck what do you think now? The paths have been updated -- is it mergeable now?

.github/workflows/build-wheel-wrapper.yml
```diff
@@ -21,5 +21,7 @@ on:
 jobs:
   python-wrapper-wheel:
     name: Python Wrapper Wheel
-    uses: ecmwf-actions/reusable-workflows/.github/workflows/python-wrapper-wheel.yml@main
```
wdeconinck, line 24: @iainrussell is this change from `ecmwf-actions/reusable-workflows` to `ecmwf/reusable-workflows` now for all our actions to be recommended?

wdeconinck (approved):
Thanks @tmi this is OK for me now.

PR #255, opened 2025-01-30
Assemble global (serial) interpolation matrix from a parallel Interpolation

sbrdar:
* add AssembleGlobalMatrix from a given atlas::Interpolation
* add unit test 'test_interpolation_mpi_global_matrix' to check that: global matrix * field == par_interpolation(par_field)
* sandbox tool 'atlas-global-matrix' with I/O for global matrices from interpolation and grids.

(with @wdeconinck )

PR #258, opened 2025-02-13
Add distribution of serial interpolation matrix

sbrdar:
1. add method 'distribute_global_matrix (src_fs, tgt_fs, global_matrix, mpi_root)' to return local interpolation matrices
2. make sure that interpolation method in Atlas which take cache matrix as an argument actually use that cache
3. extend test_interpolation_global_matrix to: 1) create a global matrix from a parallel interpolator (was there before); 2) distribute the global matrix to task-local matrices (new); 3) compare the remapped field gathered in a global field on the single task with the global-matrix remapped global field.
4. extend sandbox/atlas-global-matrix to be able to read from the disc and distribute matrix to the parallel interpolator

### DIFF
--- a/src/atlas/interpolation/AssembleGlobalMatrix.cc
+++ b/src/atlas/interpolation/AssembleGlobalMatrix.cc
@@ -201,7 +201,6 @@ void distribute_global_matrix(const linalg::SparseMatrixView<ViewValue,ViewIndex
     cols.resize(nnz_loc);
     vals.resize(nnz_loc);
 
-    size_t idx = 0;
     if (mpi_rank == mpi_root) {
         std::vector<std::vector<Index>> send_rows(mpi_size);
         std::vector<std::vector<Index>> send_cols(mpi_size);
@@ -242,10 +241,6 @@ void distribute_global_matrix(const linalg::SparseMatrixView<ViewValue,ViewIndex
 
 linalg::SparseMatrixStorage distribute_global_matrix(const FunctionSpace& src_fs, const FunctionSpace& tgt_fs, const linalg::SparseMatrixStorage& gmatrix, int mpi_root) {
     ATLAS_TRACE("distribute_global_matrix");
-    const auto src_part = array::make_view<int, 1>(src_fs.partition());
-    const auto tgt_part = array::make_view<int, 1>(tgt_fs.partition());
-    const auto tgt_ridx = array::make_indexview<int, 1>(tgt_fs.remote_index());
-
     Field field_tgt_part_glb = tgt_fs.createField(tgt_fs.partition(), option::global(mpi_root));
     ATLAS_TRACE_SCOPE("gather partition") {
         tgt_fs.gather(tgt_fs.partition(), field_tgt_part_glb);
--- a/src/atlas/interpolation/AssembleGlobalMatrix.cc
+++ b/src/atlas/interpolation/AssembleGlobalMatrix.cc
@@ -11,6 +11,8 @@
 
 #include "AssembleGlobalMatrix.h"
 
+#include <unordered_map>
+
 #include "eckit/linalg/types.h"
 
 #include "atlas/array.h"
@@ -164,6 +166,7 @@ linalg::SparseMatrixStorage assemble_global_matrix(const Interpolation& interpol
 template <typename ViewValue, typename ViewIndex, typename Value, typename Index> 
 void distribute_global_matrix(const linalg::SparseMatrixView<ViewValue,ViewIndex>& global_matrix,
     const array::Array& partition, std::vector<Index>& rows, std::vector<Index>& cols, std::vector<Value>& vals, int mpi_root) {
+    ATLAS_TRACE("distribute_global_matrix_lowlevel");
 
     const auto tgt_part_glb = array::make_view<int,1>(partition);
     
@@ -238,12 +241,15 @@ void distribute_global_matrix(const linalg::SparseMatrixView<ViewValue,ViewIndex
 }
 
 linalg::SparseMatrixStorage distribute_global_matrix(const FunctionSpace& src_fs, const FunctionSpace& tgt_fs, const linalg::SparseMatrixStorage& gmatrix, int mpi_root) {
+    ATLAS_TRACE("distribute_global_matrix");
     const auto src_part = array::make_view<int, 1>(src_fs.partition());
     const auto tgt_part = array::make_view<int, 1>(tgt_fs.partition());
     const auto tgt_ridx = array::make_indexview<int, 1>(tgt_fs.remote_index());
 
     Field field_tgt_part_glb = tgt_fs.createField(tgt_fs.partition(), option::global(mpi_root));
-    tgt_fs.gather(tgt_fs.partition(), field_tgt_part_glb);
+    ATLAS_TRACE_SCOPE("gather partition") {
+        tgt_fs.gather(tgt_fs.partition(), field_tgt_part_glb);
+    }
 
     using Index = eckit::linalg::Index;
     using Value = eckit::linalg::Scalar;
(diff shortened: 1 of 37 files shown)


wdeconinck:
Thanks @sbrdar , I have rebased this on latest develop, which now contains a much tougher ci that no longer permits errors. Please fix the errors before I review.

src/atlas/interpolation/AssembleGlobalMatrix.cc
```diff
@@ -11,15 +11,18 @@
...
 #include "AssembleGlobalMatrix.h"
 
+#include <unordered_map>
+
 #include "eckit/linalg/types.h"
 
 #include "atlas/array.h"
 #include "atlas/linalg/sparse/SparseMatrixToTriplets.h"
 #include "atlas/functionspace/StructuredColumns.h"
+#include "atlas/interpolation/Cache.h"
```
wdeconinck, line 21: This include should not be needed.

src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.h
```diff
@@ -42,6 +42,7 @@ class CubedSphereBilinear : public Method {
     using Method::do_setup;
     void do_setup(const FunctionSpace& source, const FunctionSpace& target) override;
     void do_setup(const Grid& source, const Grid& target, const Cache&) override;
+     void do_setup(const FunctionSpace& source, const FunctionSpace& target, const Cache&) override;
```
wdeconinck, line 45:
```suggestion
    void do_setup(const FunctionSpace& source, const FunctionSpace& target, const Cache&) override;
```
whitespace

src/atlas/interpolation/method/knn/KNearestNeighbours.cc
```diff
@@ -62,11 +62,19 @@ void KNearestNeighbours::do_setup(const Grid& source, const Grid& target, const
     do_setup(functionspace(source), functionspace(target));
 }
 
+void KNearestNeighbours::do_setup(const FunctionSpace& source, const FunctionSpace& target, const Cache& cache) {
+    if (interpolation::MatrixCache(cache)) {
+        setMatrix(cache);
+        source_ = source;
+        target_ = target;
+        buildPointSearchTree(source);
```
wdeconinck, line 70: Building the PointSearchTree should not be needed when the cache contains a matrix.

On the other hand when there's no matrix in the cache this sets up nothing. That needs fixing.
```c++
if(interpolation::MatrixCache(cache)) { ... } 
else { 
  THIS SITUATION
}
```

src/atlas/interpolation/method/knn/NearestNeighbour.cc
```diff
@@ -38,6 +38,15 @@ MethodBuilder<NearestNeighbour> __builder("nearest-neighbour");
 
 }  // namespace
 
+void NearestNeighbour::do_setup(const FunctionSpace& source, const FunctionSpace& target, const Cache& cache) {
+    if (interpolation::MatrixCache(cache)) {
+        setMatrix(cache);
+        source_ = source;
+        target_ = target;
+        buildPointSearchTree(source);
+    }
```
wdeconinck, line 47: Same case as for KNearestNeighbours::do_setup

src/atlas/interpolation/method/structured/StructuredInterpolation2D.tcc
```diff
@@ -264,6 +264,24 @@ void StructuredInterpolation2D<Kernel>::do_setup( const Grid& source, const Grid
...
+        target_ = target;
+        ATLAS_ASSERT(matrix().rows() == target.size());
+        ATLAS_ASSERT(matrix().cols() == source.size());
+        return;
+    }
+    if (functionspace::NodeColumns(source) && functionspace::PointCloud(target)) {
+        do_setup(source, target);
+        return;
+    }
+    ATLAS_NOTIMPLEMENTED;
```
wdeconinck, line 282: This needs also the handling of the "matrix_free" configuration option, meaning that when "matrix_free" is chosen we also need to ignore the matrix.

This `if` is not needed, and you can just call `do_setup(source, target)`

src/atlas/interpolation/method/structured/StructuredInterpolation3D.tcc
```diff
@@ -63,15 +63,30 @@ void StructuredInterpolation3D<Kernel>::do_setup( const Grid& source, const Grid
...
+        source_ = source;
+        target_ = target;
+        ATLAS_ASSERT(matrix().rows() == target.size());
+        ATLAS_ASSERT(matrix().cols() == source.size());
+        return;
+    }
+    if (functionspace::StructuredColumns(source) && functionspace::PointCloud(target)) {
+        do_setup( source, target );
+        return;
+    }
```
wdeconinck, line 87: Also here, handling of "matrix_free" configuration option.
This `if` is not needed, and you can just call `do_setup(source, target)`

src/atlas/interpolation/method/unstructured/FiniteElement.cc
```diff
@@ -83,6 +83,22 @@ void FiniteElement::do_setup(const Grid& source, const Grid& target, const Cache
...
+void FiniteElement::do_setup(const FunctionSpace& source, const FunctionSpace& target, const Cache& cache) { 
+    if (interpolation::MatrixCache(cache)) {
+        setMatrix(cache);
+        source_ = source;
+        target_ = target;
+        ATLAS_ASSERT(matrix().rows() == target.size());
+        ATLAS_ASSERT(matrix().cols() == source.size());
+        return;
+    }
+    if (functionspace::NodeColumns(source) && functionspace::PointCloud(target)) {
```
wdeconinck, line 95: This `if` is not needed, and you can just call `do_setup(source, target)`

src/atlas/interpolation/method/unstructured/UnstructuredBilinearLonLat.cc
```diff
@@ -80,6 +80,24 @@ void UnstructuredBilinearLonLat::do_setup(const Grid& source, const Grid& target
...
+    //  no halo_exchange because we don't have any halo with delaunay or 3d structured meshgenerator
+    if (interpolation::MatrixCache(cache)) {
+        setMatrix(cache);
+        source_ = source;
+        target_ = target;
+        ATLAS_ASSERT(matrix().rows() == target.size());
+        ATLAS_ASSERT(matrix().cols() == source.size());
+        return;
+    }
+    if (functionspace::NodeColumns(source) && functionspace::PointCloud(target)) {
```
wdeconinck, line 94: This `if` is not needed, and you can just call `do_setup(source, target)`

wdeconinck (changes requested):
Very nice work!
I have a few suggestions, see below.

src/atlas/interpolation/AssembleGlobalMatrix.cc
sbrdar: Removed. Thanks!

src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.h
sbrdar: Whitespace removed.

src/atlas/interpolation/method/knn/NearestNeighbour.cc
sbrdar: See above.

src/atlas/interpolation/method/structured/StructuredInterpolation2D.tcc
sbrdar: Thanks! Now, if the method is not matrix free and there is a cache matrix, we initialise from the cache. In all other cases do_setup(source, target)

src/atlas/interpolation/method/unstructured/FiniteElement.cc
sbrdar: Thanks! Done.

src/atlas/interpolation/method/unstructured/UnstructuredBilinearLonLat.cc
sbrdar: Yes, see above. Many thanks for reviewing!!

wdeconinck (approved):

wdeconinck:
> Question: does this PR allow to build a serial-numbered matrix directly (outside an MPI environment)?

This is PR is the inverse, it rebuilds a parallel-numbered matrix from a serial-numbered matrix. A previous PR https://github.com/ecmwf/atlas/pull/255 is building the serial-numbered matrix.

pmaciel:
> This is PR is the inverse, it rebuilds a parallel-numbered matrix from a serial-numbered matrix. A previous PR #255 is building the serial-numbered matrix.

Yes I'm aware, but at least the PR #255 testing requires MPI. So this isn't yet covered?

wdeconinck:
> Yes I'm aware, but at least the PR #255 testing requires MPI. So this isn't yet covered?

Good point. It should work as well with 1 MPI task as that is only one specific case of ANY number of MPI tasks, but a test needs to be added for that (too easy to say it "should work").

@sbrdar could you please 🙏 create a PR that adds such test making sure that we can create the serial-numbered matrix with only one MPI task, for use with "grid-sized" vectors for interpolation?

sbrdar:
@wdeconinck I have tested that it works with 1 MPI task. In addition, I compiled Atlas with HAVE_MPI=0 and the sandbox tool 'atlas-global-matrix' works there as well. The sandbox tool uses 'grid-sized' vector for interpolation, see [.](https://github.com/ecmwf/atlas/blob/4f8826bf3d9001be2294a87e2620be7075b6111b/src/sandbox/interpolation/atlas-global-matrix.cc#L246)
Do you still want a separate unit test with 1 MPI?

pmaciel:
> @wdeconinck I have tested that it works with 1 MPI task. In addition, I compiled Atlas with HAVE_MPI=0 and the sandbox tool 'atlas-global-matrix' works there as well. The sandbox tool uses 'grid-sized' vector for interpolation, see [.](https://github.com/ecmwf/atlas/blob/4f8826bf3d9001be2294a87e2620be7075b6111b/src/sandbox/interpolation/atlas-global-matrix.cc#L246) Do you still want a separate unit test with 1 MPI?

Ideally, even with eckit_HAVE_MPI=0 :-)

sbrdar:
> > @wdeconinck I have tested that it works with 1 MPI task. In addition, I compiled Atlas with HAVE_MPI=0 and the sandbox tool 'atlas-global-matrix' works there as well. The sandbox tool uses 'grid-sized' vector for interpolation, see [.](https://github.com/ecmwf/atlas/blob/4f8826bf3d9001be2294a87e2620be7075b6111b/src/sandbox/interpolation/atlas-global-matrix.cc#L246) Do you still want a separate unit test with 1 MPI?
> 
> Ideally, even with eckit_HAVE_MPI=0 :-)

Yes, even that is tested. I used atlas-bundle with '--without-mpi'.

wdeconinck:
> Ideally, even with eckit_HAVE_MPI=0 :-)

Yes atlas only uses eckit MPI. What @sbrdar means is with MPI disabled via eckit indeed, which happens with the `--without-mpi` option in atlas-bundle.

Please make sure there is a test without `MPI` argument  in `ecbuild_add_test`, so that this gets tested in e.g. ecmwf-toolbox which has no MPI enabled.

Issue #260, opened 2025-02-28
Grid getter in FunctionSpace


tom-j-h:
### Is your feature request related to a problem? Please describe.

I'm working on something where it would be helpful for a FunctionSpace to be able to return a Grid.

If I understand correctly, all FunctionSpaces (excluding Spectral) should have a Grid underlying them in some way, whether that's directly ((Block)StructuredColumns), via the Mesh (Cell/Edge/NodeColumns), or implicitly via a list of Points (PointCloud).

### Describe the solution you'd like

Could we have a getter in FunctionSpace, with implementations in each of the individual types except Spectral? This should be simple for the xColumns FunctionSpaces. PointCloud is trickier - I think we could generate a Grid on the fly using iterate() and return that, or return a Grid that was copied on construction of the PointCloud if it was constructed with one.

### Describe alternatives you've considered

At the moment, I think I'd have to dynamic_cast to get the FunctionSpace of particualr type from the base FunctionSpace, then get Grids in the same way as described above.

### Additional context

My use case is that we are implementing a simple parallel NetCDF I/O system for FieldSets in JEDI-OOPS. The first step before writing is to redistribute the data (using Redistribution) to a FunctionSpace on the same Grid as the native Fields (hence the need to obtain the grid) but who's Distribution is simply putting ~equal numbers of points onto each PE in order of Atlas's global index. The data is then written to the file in that order so that the files are agnostic to the number of PEs/domain decomposition of the native Fields. For reading the data back in, we do the reverse.

### Organisation

Met Office

tom-j-h:
I've just updated the title and description a bit.

wdeconinck:
I think this would be a good change and one I had been considering for a while.
The one concern I had, as you pointed out were the FunctionSpaces that have no attached grid.

But also as you pointed out these could be created except for the Spectral.
- PointCloud
- Mesh-based such as NodeColumns, CellColumns. It is not guaranteed that these have an attached grid, if they have just been created by a mesh reader or other means. In that case the same suggested procedure as for PointCloud could be done and a unstructured grid would need to be created.

tom-j-h:
Thanks for your comment, I wasn't aware that those types may need to construct a Grid too.

@mo-jonasganderton is also interested in adding this functionality and has some ideas on how to improve the PR. We can let you know when it's ready for review.

tom-j-h:
I've just made some updates to the PR.

wdeconinck:
[...]

PR #264, opened 2025-03-13
Add Grid getter to FunctionSpace

tom-j-h:
See #260 for motivation.

I've used the name get_grid_copy(), since BlockStructuredColumns and StructuredColumns already have a grid() method and this would cause a conflict.

As suggested in the Issue, PointCloud stores a Grid if constructed with one. It returns this if it has it, otherwise it generates one using its own iterator. I've added a test case that checks both paths. Although the test is apparently run with MPI=2, the (nb_partitions_ == 1) branch in the added function is being used to generate the Grid because the PointCloud constructor that takes a vector of points does not set nb_partitions_. I've been using the code externally and the parallel branch is giving me correct results, but I think ideally this should be tested. Not sure how to resolve. Should that constructor have a nb_partitions_ = comm.size() statement?


tom-j-h:
I've updated the code. Changed name of method to base_grid (suggested by @mo-jonasganderton), implemented generators in mesh-based FunctionSpaces, and added test for all FunctionSpace types. If a Grid is available during FunctionSpace construction or it is generated then this is stored as a member variable in case it's requested again. There's quite a bit of repeated code for the Grid generation, but I wasn't sure where it should be factored out to.

wdeconinck:
Thanks @tom-j-h for this nice work! It will be very useful.

> I've used the name get_grid_copy(), since BlockStructuredColumns and StructuredColumns already have a grid() method and this would cause a conflict.

> Changed name of method to base_grid 

The `FunctionSpace::base_grid()` and `StructuredColumns::grid()`  should be identical in fact. I think we could safely use the same name, i.e. `FunctionSpace::grid()`, unless you have another case in mind?

tom-j-h:
> The `FunctionSpace::base_grid()` and `StructuredColumns::grid()` should be identical in fact. I think we could safely use the same name, i.e. `FunctionSpace::grid()`, unless you have another case in mind?

Ah, you're right! Since we're now constructing a Grid if there isn't one and storing it as a member variable, we can always return a const reference. So we can change the name of the virtual method to just "grid" and BlockStructuredColumns and StructuredColumns can override the function with their existing methods (thanks to covariant return types, which I just learnt about). Thanks. Changes committed.

wdeconinck (approved):
This is looking perfect! 👌
Thanks @tom-j-h ! 🙏

wdeconinck (changes requested):
Please fix the compilation errors flagged by CI.
FYI, the CI also compiles with `-DENABLE_WARNINGS=ON -DENABLE_WARNING_AS_ERROR=ON` in order to avoid any warnings creeping in.

tom-j-h:
Thanks very much! I think this should be okay now, CI running...

wdeconinck (approved):
Great!

PR #267, opened 2025-03-20
Add offloading of discontiguous arrays

wdeconinck:
This allows to offload existing wrapped data which may discontiguous as a slice of a larger array. This is e.g. the case when using a Multifield.

It makes use of HIC.

This is co-developed with @sbrdar

### DIFF
--- a/src/atlas/array/native/NativeDataStore.h
+++ b/src/atlas/array/native/NativeDataStore.h
@@ -309,12 +309,52 @@ private:
 template <typename Value>
 class WrappedDataStore : public ArrayDataStore {
 public:
-    WrappedDataStore(Value* host_data, size_t size): size_(size), host_data_(host_data) {
+
+    void init_device() {
         if (ATLAS_HAVE_GPU && devices()) {
             device_updated_ = false;
         }
         else {
             device_data_ = host_data_;
+        }
+    }
+
+    WrappedDataStore(Value* host_data, size_t size): host_data_(host_data), size_(size) {
+        init_device();
+    }
+
+    WrappedDataStore(Value* host_data, const ArraySpec& spec):
+        host_data_(host_data),
+        size_(spec.size())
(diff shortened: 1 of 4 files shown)


src/tests/field/fctest_field_wrap_gpu.F90
```diff
@@ -37,6 +37,54 @@ module fcta_Field_wrap_device_fixture
 
 ! -----------------------------------------------------------------------------
 
+subroutine kernel_1(view, N)
+  real(c_double) :: view(:,:,:)
```
sbrdar, line 41: I would suggest to leave a comment in the same line that because this is a kernel launcher function, we should not have the 'intent' keyword for 'view' according to the OpenACC standard. These are easy-to-omit things, but helpful to know.

sbrdar (approved):

src/tests/field/fctest_field_wrap_gpu.F90
wdeconinck: Does this mean even `intent(inout)` is not OK? That is usually implied I thought.

src/atlas/array/native/NativeDataStore.h
```diff
@@ -318,6 +319,45 @@ class WrappedDataStore : public ArrayDataStore {
...
+    WrappedDataStore(Value* host_data, const ArraySpec& spec):
+        host_data_(host_data),
+        size_(spec.size())
+    {
+        init_device();
+        contiguous_ = spec.contiguous();
+        if (! contiguous_) {
+            int break_idx = 0;
+            size_t shp_mult_rhs = spec.shape()[spec.rank() - 1];
+            for (int i = spec.rank() - 1; i > 0; i--) {
```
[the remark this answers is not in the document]
wdeconinck: That is a good idea.

PR #268, opened 2025-03-21
Add device offload operators on FieldSet

wdeconinck:
This work by @sbrdar allows to handily offload an entire fieldset or a subset of contained fields referenced by name or indices using initialiser lists.
Thanks @sbrdar !

### DIFF
--- a/src/tests/field/test_field_acc.cc
+++ b/src/tests/field/test_field_acc.cc
@@ -98,20 +98,22 @@ CASE("test_fieldset_acc") {
     fieldset.add(mfield);
     fieldset.add(Field("mask", make_datatype<int>(), array::make_shape(10)));
 
-    for (auto f : fieldset) {
-        f.allocateDevice();
-    }
+    fieldset.allocateDevice({"temperature"});
 
-    auto field = fieldset.field(1);
+    auto field = fieldset.field(0);
     auto view = array::make_view<double, 2>(field);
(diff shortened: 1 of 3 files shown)


src/tests/field/test_field_acc.cc
```diff
@@ -98,28 +98,29 @@ CASE("test_fieldset_acc") {
...
 // TODO: gridtools storage does not implement view.index(...) at the moment
 
+    auto dview = array::make_device_view<double, 2>(field);
     double* dptr = dview.data();
 #pragma acc parallel deviceptr(dptr)
     {
         double t = dptr[dview.index(3,2)];
         dptr[dview.index(3,2)] = 1. + t;
     }
     field.updateHost();
```
sbrdar, line 123: For the code coverage, one could also use "fieldset.updateHost()" which only updates the host for device-allocated variables.

src/tests/field/test_field_acc.cc
```diff
@@ -98,28 +98,29 @@ CASE("test_fieldset_acc") {
...
-
     EXPECT_EQ( view(3,2), 2. );
```
sbrdar, line 125: Also for the code coverage, we could call fieldset.deallocateDevice() here.

sbrdar (approved):
Thank you @wdeconinck for this PR. I approve it and also add two suggestions for improving the code coverage.

PR #273, opened 2025-04-01
Use pluto default memory resources in atlas::array

wdeconinck:
NOTE, this PR applies only to `atlas::array::Array`, which also underpins `atlas::Field`!

Until now atlas was using a fixed strategy for host and device allocations in the atlas::array::Array
- host -> At least 64-byte aligned memory allocation, via `posix_memalign` (before #270) or 256-byte aligned `pluto::host_resource()` after.
- device -> `cudaMalloc` or `hipMalloc` via `hicMalloc` (before #270) or `pluto::device_resource()` after.

With this PR we now use:
- host -> `pluto::host::get_default_resource()`
- device -> `pluto::device::get_default_resource()` 

### Example use case 1
Suppose we want to use a pool of pinned memory for the host, and a pool of device memory for device. This makes for fast host-device memory transfers.

We can setup this up with C++:
 - Using `pluto::memory_resource*` argument: 
   ```c++
   pluto::host::set_default_resource(pluto::pinned_pool_resource());
   pluto::device::set_default_resource(pluto::device_pool_resource());
   ```
 - Using string argument to registered names:
    ```c++
    pluto::host::set_default_resource("pluto::pinned_pool_resource");
    pluto::device::set_default_resource("pluto::device_pool_resource()");
    ```
 
 Or with Fortran:
 * Using `type(pluto_memory_resource)` argument: 
    ```Fortran 
    call pluto%host%set_default_resource(pluto%pinned_pool_resource())
    call pluto%device%set_default_resource(pluto%device_pool_resource())
    ````
 - Using string argument to registered names:
    ```Fortran 
    call pluto%host%set_default_resource("pluto::pinned_pool_resource")
    call pluto%device%set_default_resource("pluto::device_pool_resource")
    ````

Initial values for `pluto::host::get_default_resource()` and `pluto::device::get_default_resource()` are respectively `pluto::host_resource()` and `pluto::device_resource()`.
These initial values are also configurable via environment variables:
  `export PLUTO_HOST_MEMORY_RESOURCE=pluto::pinned_pool_resource`
  `export PLUTO_DEVICE_MEMORY_RESOURCE=pluto::device_pool_resource`

Pluto provides following memory_resources, accessible by name or function:
- `pluto::host_resource`,  `pluto::host_pool_resource`
- `pluto::pinned_resource`, `pluto::pinned_pool_resource`
- `pluto::managed_resource`, `pluto::managed_pool_resource`
- `pluto::device_resource`, `pluto::device_pool_resource`

### Example use case 2
Using zero-copy approach or managed memory for device:
When following is retrieved for the `pluto::device::get_default_resource()`
-  `pluto::pinned_resource`, `pluto::pinned_pool_resource`
- `pluto::managed_resource`, `pluto::managed_pool_resource`

Then we assume the memory access for host and device is shared (a.k.a. unified): 
- The host-data will be allocated using the configured device memory_resource
- The device data will point to the host-data.

This PR also adds a `label` to the atlas::Array creation via `Field`, and passes this to pluto for tracing reasons.

### DIFF
--- a/src/atlas/array/native/NativeDataStore.h
+++ b/src/atlas/array/native/NativeDataStore.h
@@ -11,20 +11,17 @@
 #pragma once
 
 #include <algorithm>  // std::fill
-#include <atomic>
-#include <cstdlib>  // posix_memalign
 #include <limits>   // std::numeric_limits<T>::signaling_NaN
-#include <sstream>
+#include <string>
+#include <iostream>
 
 #include "pluto/pluto.h"
 
 #include "atlas/array/ArrayDataStore.h"
-#include "atlas/library/Library.h"
+#include "atlas/array/ArraySpec.h"
 #include "atlas/library/config.h"
 #include "atlas/parallel/acc/acc.h"
 #include "atlas/runtime/Exception.h"
-#include "atlas/runtime/Log.h"
-#include "eckit/log/Bytes.h"
 
 #define ATLAS_ACC_DEBUG 0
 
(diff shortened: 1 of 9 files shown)


src/atlas/array/native/NativeDataStore.h
```diff
@@ -93,12 +49,61 @@ template <typename Value>
 void initialise(Value[], size_t) {}
 #endif
 
+static bool test_device_is_shared() {
+    // This is currently a bit hacky, to be improved!
+    auto& device_resource = *pluto::device::get_default_resource();
```
sbrdar, line 54: Would this work without going to pointers? Arguably, device_resource ==  pluto::pinned_resource() seems most intuitive for the comparisons in the next lines of code.

sbrdar (approved):
Many thanks for this PR, @wdeconinck. Very useful and important to consolidate the GPU offloading in Atlas. Only one comment to address.

src/atlas/array/native/NativeDataStore.h
wdeconinck: In fact this is without pointers, dereferencing it.
It can be seen as:
```c++
memory_resource& device_resource = *pluto::device::get_default_resource();
```
What you're suggesting is pointer comparison. That could also work but not generally.

PR #275, opened 2025-04-22
Update interpolation to support hicsparse backend

l90lpa:
This PR modifies interpolation code to enable support for the `hicsparse` linalg backend when using matrix based interpolation methods. It should be considered as a first draft as I think there are a number of areas that one might want to change. Here are some of my thoughts and modifications contained in this PR:
- Since this PR adds a backend that performers its computation in the device memory space, I think there is a question regarding what the user's responsibility is to regarding providing inputs in the same memory space as the desired backend (here I put desired, because in some code paths a user might opt for the hicsparse backend but the openmp backend might be used instead). As an initial choice I decided to allow the user to provide inputs in any memory space, but provide no guarantee about what memory space the results might be returned - hence leaving it to the user to move data after the operation if they have to or want to as this felt like it might make Method.cc less error prone, and reduce the number of essential data movement operations. That said, I can't say that I am entirely pleased with this choice.
- As there are know multiple memory spaces on which a linalg backend might operate, I currently have a helper function at the top of Method.cc to help with knowing where to place data, but I think this might be best moved onto the `Backend` object in `linalg/sparse`.
-  I have also added a few helper functions at the top of Method.cc for getting array views that have up-to-date data in the required memory space. If these seem broadly useful it might be worthing moving them out. Note, I didn't use the existing `sync`
method because this can introduce unnecessary data movements.
- I have added a couple of tests by copying some existing interpolation test and slightly reducing/modifying them. That said, I'd appricate advise if you think there are better tests that I could put in.
- I have modified Method.cc so as to be able to make use of the `on_device` halo exchange.


src/atlas/interpolation/method/Method.cc
```diff
@@ -95,34 +95,127 @@ void set_missing_values(Field& tgt, const std::vector<idx_t>& missing) {
...
 template <typename Value>
 void Method::interpolate_field_rank1(const Field& src, Field& tgt, const Matrix& W) const {
-    auto backend = std::is_same<Value, float>::value ? sparse::backend::openmp() : sparse::Backend{linalg_backend_};
-    auto src_v   = array::make_view<Value, 1>(src);
-    auto tgt_v   = array::make_view<Value, 1>(tgt);
+    auto backend = sparse::Backend{linalg_backend_};
+    
+    if (backend.type() == "hicsparse" && !std::is_same<eckit::linalg::Scalar, Value>::value) {
+        ATLAS_NOTIMPLEMENTED; // hicsparse does not support mixed double-float
+    }
```
wdeconinck, line 176: I believe we still also need
```c++

    // Switch to OpenMP when using eckit_linalg and using float
    if (backend.type() == "eckit_linalg" && !std::is_same<eckit::linalg::Scalar, Value>::value) {
        backend = sparse::backend::openmp();
    }
```

src/atlas/interpolation/method/Method.cc
```diff
@@ -144,54 +237,109 @@ void Method::interpolate_field_rank2(const Field& src, Field& tgt, const Matrix&
...
 }
 
 template <typename Value>
 void Method::adjoint_interpolate_field_rank1(Field& src, const Field& tgt, const Matrix& W) const {
-    auto backend = std::is_same<Value, float>::value ? sparse::backend::openmp() : sparse::Backend{linalg_backend_};
+    auto backend = sparse::Backend{linalg_backend_};
+    
+    if (backend.type() == "hicsparse" && !std::is_same<eckit::linalg::Scalar, Value>::value) {
+        ATLAS_NOTIMPLEMENTED; // hicsparse does not support mixed double-float
+    }
```
wdeconinck, line 289: Also here:
```c++

    // Switch to OpenMP when using eckit_linalg and using float
    if (backend.type() == "eckit_linalg" && !std::is_same<eckit::linalg::Scalar, Value>::value) {
        backend = sparse::backend::openmp();
    }
```

wdeconinck (changes requested):
Thanks @l90lpa this is looking very good now.
Please take care of following suggestions.

src/atlas/interpolation/method/Method.cc
l90lpa: Hi @wdeconinck, thanks for catching this! I've made these changes. I've also made the logic for switching to the OpenMP backend in the case of rank 3 fields more explicit to match the pattern in the rest of the functions.

wdeconinck (approved):
All looks good now. I hope you find the changes OK, @l90lpa

PR #287, opened 2025-05-23
Move shebag to begining of c2f.py

Ozaq:
According to execve(2) interpreter scripts must start with #! on the first line.


wdeconinck (approved):
That's great. I'm wondering how this could stay under the radar for so long.
Interested to hear which environment was used (cmake, compilers, python, shell).

PR #293, opened 2025-06-06
Added OpenMP to the cubed-sphere interpolator and the matching cubed sphere partitioner.

odlomax:
This adds some OpenMP that was skipped in the initial implementation. Similar in intent to #292 

I'd like to eventually retire this interpolator in favour of finite-element interpolation, but we're not quite ready yet.

I've also removed a nasty metadata code smell which was needed by the old cubed-sphere wind interpolation scheme. I've accordingly removed the test that used the metadata.

### DIFF
--- a/.github/ci-hpc-config.yml
+++ b/.github/ci-hpc-config.yml
@@ -7,7 +7,8 @@ mpi_on:
     modules:
       - ninja
     modules_package:
-      - atlas:fftw,eigen,openmpi
+      - atlas:fftw,eigen,qhull,python3/3.12,openmpi
+      - fckit:python3/3.12,openmpi
       - eckit:openmpi
     dependencies:
       - ecmwf/ecbuild@develop
@@ -15,13 +16,16 @@ mpi_on:
       - ecmwf/fckit@develop
     parallel: 64
     ntasks: 16
+    env:
+      - ATLAS_FPE=0
 
 mpi_off:
   build:
     modules:
       - ninja
     modules_package:
-      - atlas:fftw,eigen
+      - atlas:fftw,eigen,qhull,python3/3.12
+      - fckit:python3/3.12
     dependencies:
       - ecmwf/ecbuild@develop
       - ecmwf/eckit@develop
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -93,6 +93,7 @@ jobs:
                   - -DENABLE_ECKIT_SQL=OFF
                   - -DENABLE_ECKIT_CMD=OFF
                   - -DENABLE_AIO=OFF
+                  - -DENABLE_ECKIT_GEO=OFF
                   - -DENABLE_WARNINGS=OFF
                   - ${{ matrix.cmake_options || '' }}
               ecmwf/fckit:
--- a/pluto/src/pluto_f/pluto_module.F90
+++ b/pluto/src/pluto_f/pluto_module.F90
@@ -7,6 +7,10 @@ private
 public :: pluto, pluto_memory_resource, pluto_allocator
 
 interface
+    function c_pluto_devices() result(devices) bind(c)
+        use iso_c_binding, only: c_int
+        integer(c_int) :: devices
+    end function
     subroutine c_pluto_host_set_default_resource_name(name, name_size) bind(c)
         use iso_c_binding, only: c_ptr, c_int
         type(c_ptr), value, intent(in) :: name
@@ -249,6 +253,7 @@ type pluto_t
     type(pluto_scope_t)  :: scope
     type(pluto_trace_t)  :: trace
 contains
+    procedure, nopass :: devices => pluto_devices
     procedure, nopass :: get_registered_resource => pluto_get_registered_resource
     procedure, nopass :: new_delete_resource => pluto_new_delete_resource
     procedure, nopass :: null_memory_resource => pluto_null_memory_resource
@@ -326,6 +331,11 @@ function pluto_trace_enabled()
     else
         pluto_trace_enabled = .false.
     endif
+end function
+
+function pluto_devices()
+    integer(c_int) :: pluto_devices
+    pluto_devices = c_pluto_devices()
 end function
 
 function pluto_get_registered_resource(name) result(memory_resource)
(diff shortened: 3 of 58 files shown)


src/atlas/grid/detail/partitioner/MatchingMeshPartitionerCubedSphere.cc
```diff
@@ -31,7 +32,7 @@ void MatchingMeshPartitionerCubedSphere::partition(const Grid& grid, int partiti
 
     // Loop over grid and set partioning[].
     auto lonlatIt = grid.lonlat().begin();
-    for (gidx_t i = 0; i < grid.size(); ++i) {
+    atlas_omp_parallel_for(gidx_t i = 0; i < grid.size(); ++i) {
```
wdeconinck, line 35: This will incur a race condition on the `lonlatIt` iterator.
Probably what you want is this (not tested):
```c++
const size_t num_threads = atlas_omp_get_max_threads();
const size_t size        = grid.size();
atlas_omp_parallel {
    // This is now executed concurrently in each available OpenMP threads
    // All thread-private data here:
    const size_t thread_num   = atlas_omp_get_thread_num();
    const size_t thread_begin = (thread_num * size) / num_threads;
    const size_t thread_end   = (thread_num + 1) * size) / num_threads;
    auto lonlatIt = grid.lonlat().begin() + thread_begin; // Move iterator to thread_begin
    for (size_t i = thread_begin; i<thread_end; ++i, ++lonlatIt) { // regular for, already within an OpenMP thread
        const auto& lonlat = *lonlatIt;
        partitioning[i] = finder.getCell(lonlat, listSize, edgeEpsilon, epsilon).isect ? mpi_rank : -1;
    }
}

File: src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.cc [hunk 2]
```diff
@@ -75,24 +75,21 @@ void CubedSphereBilinear::do_setup(const FunctionSpace& source, const FunctionSp
                     std::to_string(i) + ".");
             }
 
-            tileIndex.push_back(tijView(cell.idx, 0));
```
wdeconinck, line 78: tileIndex is not used elsewhere? dead code?

wdeconinck (commented):
Thanks @odlomax this is a good idea. I have a few suggestions below.

Also I'm wondering why the complexity of the variant and visitor. You can directly insert into an oversized triplet vector, like done in #292 . The allocated memory is anyway the same because the variant takes the size of the largest type. Then you also avoid the extra visitor and copy into a new vector of triplets. The eckit::SparseMatrix constructor skips over empty triplets during construction.

src/atlas/grid/detail/partitioner/MatchingMeshPartitionerCubedSphere.cc
odlomax: Whoops!

src/atlas/interpolation/method/cubedsphere/CubedSphereBilinear.cc
odlomax: It's used in our model interface, but I plan on removing the code that before we advance our current Atlas tag.

odlomax:
> Thanks @odlomax this is a good idea. I have a few suggestions below.
> 
> Also I'm wondering why the complexity of the variant and visitor. You can directly insert into an oversized triplet vector, like done in #292 . The allocated memory is anyway the same because the variant takes the size of the largest type. Then you also avoid the extra visitor and copy into a new vector of triplets. The eckit::SparseMatrix constructor skips over empty triplets during construction.

Ah, I forgot about the triplet skipping! I'll do that then.

src/atlas/grid/detail/partitioner/MatchingMeshPartitionerCubedSphere.cc
odlomax: The lon-lat iterator is random access, so `const auto& lonLat = *(lonlatIt + i)` should work, right? Feels a bit tidier if I can avoid direct omp library calls.

odlomax: Yikes. Consider me educated.

odlomax: More worrying, if we're having this debate, then the affected code is clearly untested...
I'd better sort that out!

src/tests/interpolation/test_interpolation_cubedsphere.cc
```diff
@@ -95,6 +69,66 @@ double dotProd(const Field& a, const Field& b) {
...
+        // This does not dangle, because C++ makes no sense!
+        const auto& lonLat = *(begin + i);
+        const auto refLonLat = *iterator;
+        ++iterator;
+
+        EXPECT_APPROX_EQ(lonLat.lon(), refLonLat.lon(), 1e-14);
+        EXPECT_APPROX_EQ(lonLat.lat(), refLonLat.lat(), 1e-14);
+
+        // Only now, *(begin + i), do you have my permission to die.
+    }
```
odlomax, line 89: So, funny thing, const lvalue refs to temporary object members can extend the lifetime of the host object...

wdeconinck: This seems like the worst code practice and should not be relied on for sure :)

odlomax:
I've removed the redundant headers and the spooky C++ test.

I went with the `const auto lonLat = *(lonlatIt + i)` assignment in the end. The presence of the `+` operator on the iterator expresses intent that it should be random access. I think `std::random_access_iterator_tag` is actually set someone underneath the layers.

I'm happy to go with the other solution, but it seems like it just moves the complexity from an odd dereferencing pattern to more explicit thread management.

odlomax:
I present option 4, because why not? 🙃

odlomax:
> > I present option 4, because why not? 🙃
> 
> I'm still not vouching for this solution, but if you're happy with the performance, and it's an improvement over the non-threaded version then all good for me.

There's time for it to go wrong yet! nvhpc hates lambdas with OpenMP!

odlomax:
> So my understanding with this proposal (4), is that essentially this iterates sequentially, stores it in a vector of points as unstructured grid, and then uses OpenMP to compute the partitioning, using faster random access in the vector.

Yeah, that's the gist of it!

I don't think we ever actually use this partitioner on other cubed spheres. I didn't realise (but should have expected) that the cubed sphere iterator doesn't do random access properly.

PR #294, opened 2025-06-06
Add a vector_component option utility

odlomax:
This PR add an `option::vector_component` utility to manage the vector field component meta data required by `pack_vector_fields` and the spherical vector interpolation class.

This sets the metadata at field creation. A user should not need to modify it afterwards.

### DIFF
--- a/src/atlas/functionspace/PointCloud.cc
+++ b/src/atlas/functionspace/PointCloud.cc
@@ -513,7 +513,7 @@ void PointCloud::set_field_metadata(const eckit::Configuration& config, Field& f
         field.metadata().set("type", config.getString("type"));
     }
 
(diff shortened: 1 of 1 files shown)


src/atlas/functionspace/PointCloud.cc
```diff
@@ -512,6 +512,10 @@ void PointCloud::set_field_metadata(const eckit::Configuration& config, Field& f
     if (config.has("type")) {
         field.metadata().set("type", config.getString("type"));
     }
+
+        if (config.has("vector_component")) {
```
wdeconinck, line 516:
```suggestion
    if (config.has("vector_component")) {
```

wdeconinck (approved):
Nice @odlomax !

PR #302, opened 2025-07-03
[python] add fftw dependency

tmi:
just including fftw as a dependency in the python wheels


pmaciel:
@tmi @wdeconinck This is an important change :-)

tmi:
note: converted back to `Draft` -- technically is ok, but legal aspects unclear

wdeconinck:
It has been decided at ECMWF not to use FFTW in the python wheels, but instead rely on the pocketfft backend introduced with PR #308 for this purpose.

wdeconinck:
> updated the PR, tested it still works https://github.com/ecmwf/python-develop-bundle/actions/runs/17762319308/job/50477752299

It seems I don't have permissions for this repository, 404 errors.

wdeconinck (approved):
Good to merge!

PR #305, opened 2025-07-21
Update MeshBuilder API

wdeconinck:
Updates to the MeshBuilder and TriangularMeshBuilder API

- Changes to allow to use arbitrary indexing base for global_index (e.g. 0-based or 1-based) to define nodes, elements and connectivities.

- API that uses mdspan

- Fixes to allow strided node-coordinates

- Allow Fortran array slices for (strided) coordinates in TriangularMeshBuilder

@fmahebert this may be affecting you. Previous APIs are still working but are deprecated, with a compiler warning.
Just adding a `global_index_base` argument should remove the warning.
 

https://sites.ecmwf.int/docs/atlas/static-analyzer/PR-305

### DIFF
--- a/src/atlas/mesh/MeshBuilder.cc
+++ b/src/atlas/mesh/MeshBuilder.cc
@@ -25,18 +25,18 @@
 namespace atlas {
 
 namespace detail {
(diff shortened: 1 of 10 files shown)


fmahebert:
Thanks for the ping on this @wdeconinck !

Issue #309, opened 2025-08-14
Empty adjoint matrix in execute_adjoint when using SphericalVector interpolator


l90lpa:
### What happened?

Hi @wdeconinck, I recently ran into an assert in `Interpolation::execute_adjoint` because of an empty adjoint matrix (`Method::matrix_transpose_`) when using the `SphericalVector` interpolator. The issue seems to be related to the use of `Method::setMatrix` inside `SphericalVector::do_setup` and the code that was added to `Method.cc` for the new sparse matrix storage.

On constuction of a `SphericalVector` interpolator `Method::setup` is called which in turn calls the `SphericalVector::do_setup`. Inside `SphericalVector::do_setup` it creates an interpolation object for the scalar field scheme and then uses `Method::setMatrix(interpolation::MatrixCache matrix_cache)` (making use of the conversion constructor `MatrixCache(Interpolation&)`). This particular `setMatrix` has the affect of nullifying the member `Method::matrix_shared_`. Now when control returns to `Method::setup`, on completion of `SphericalVector::do_setup`. it does some checks to determine if it should construct an adjoint matrix. One of these checks is `matrixAllocated()` which returns false because `Method::matrix_shared_` is null, however, there is a matrix in `Method::matrix_cache_`. Because `matrixAllocated()` returns false, an adjoint matrix isn't created, which results in an assert failing when calling `Interpolation::execute_adjoint`. I'm not sure if the issue is that, `matrixAllocated()` doesn't check enough places, or that the `setMatrix` call in `SphericalVector::do_setup` isn't appropriate.

### What are the steps to reproduce the bug?

Build a `SphericalVector` interpolator with adjoint enabled and call `execute_adjoint`.

### Version

latest

### Platform (OS and architecture)

Ubuntu 22.04

### Relevant log output

```shell

```

### Organisation

JCSDA

l90lpa:
Hi @wdeconinck, I'm just wondering if you've had a chance to look at this issue, because I think we are dependent on a fix for this before JEDI can move to the latest version of Atlas.

wdeconinck:
Hi @l90lpa thank you for this. I did not get the chance to look yet, but I will try to reproduce and let you know.

l90lpa:
[...]

PR #314, opened 2025-09-03
Fix empty adjoint matrix for SphericalVector for scalar fields

wdeconinck:
This PR fixes issue #309  where an empty adjoint matrix occurs for handling scalar fields with the SphericalVector interpolation method.

@l90lpa please verify this is OK.


wdeconinck:
I need to fix a few bugs more it seems 😅

wdeconinck:
Thanks @odlomax it is now sorted, and this PR is ready to merge. Failing tests now are only downstream and have nothing to do with this PR it seems.

I'm not sure it was working before as intended for adjoints of scalar fields. At least the test @l90lpa created was not there.

PR #317, opened 2025-09-08
Add Locator to find at which partition and at which index a global index is located

wdeconinck:
Provide a communication pattern helper that, given global indices, find the partition and remote index of that global index.
 

https://sites.ecmwf.int/docs/atlas/static-analyzer/PR-317


twsearle:
So happy to see this! I was wanting to find the time to try and do this for years! :)

sbrdar (approved):
Many thanks for this valuable feature @wdeconinck ! Very useful for further developments.

Issue #319, opened 2025-09-18
Cannot find pocketfft on non-Internet machine


DJDavies2:
### What happened?

I'm getting this cmake failure:

CMake Error at atlas/CMakeLists.txt:106 (file):
  file RENAME failed to rename

    /home/users/david.davies/cylc-run/mo-bundle-build-stack-415/share/make_mo_base__ex_gnu/atlas/download/pocketfft-0fa0ef591e38c2758e3184c6c23e497b9f732ffa

  to

    /home/users/david.davies/cylc-run/mo-bundle-build-stack-415/share/make_mo_base__ex_gnu/atlas/pocketfft

  because: No such file or directory

I presume this is because the machine it is running on isn't connected to the Internet.  Is there a solution for this?  

### What are the steps to reproduce the bug?

Run cmake without an internet connection.

### Version

develop

### Platform (OS and architecture)

Linux not on Internet

### Relevant log output

```shell

```

### Organisation

Met Office

wdeconinck:
I have implemented above mentioned option 2 in #320.
Please verify that this solves the problem.

DJDavies2:
Thanks, this seems to have worked for me.

wdeconinck:
Great, merged in develop now.

PR #320, opened 2025-09-19
Make FFT backend pocketfft optional

wdeconinck:
This PR adds a feature POCKETFFT that gets disabled when pocketfft cannot be downloaded on non-internet machines.
It is to be noted that when both pocketfft and FFTW are not enabled, then Atlas cannot perform inverse spherical harmonics transforms with the TransLocal implementation.
This fixes issue #319 

https://sites.ecmwf.int/docs/atlas/static-analyzer/PR-320


DJDavies2:
I tested this here on platforms connected to the Internet and platforms that worked; it seemed to work either way, so this seems to fix the problem from my point of view.

Issue #328, opened 2025-11-05
Modern interpolation methods for unstructured grids.


lewisn-met:
### Is your feature request related to a problem? Please describe.

While the existing [finite element](https://github.com/ecmwf/atlas/blob/develop/src/atlas/interpolation/method/unstructured/FiniteElement.cc) interpolation method for unstructured grids implemented in Atlas is sufficiently accurate for most use cases, the algorithm chosen has a number of downsides, namely,

- It is prone to C1 discontinuity: The approach to interpolating on quads involves dividing the quad into two triangles, however, the result of the interpolation is generally not smooth and of lower accuracy along the dividing line. 
- It is not invariant under cycling of polygon vertices: if the mesh is read in a different order, the result of the interpolation may change, generating a reproducibility issue. Furthermore, the orientation of the C1 discontinuity varies under this cycling, so the difference can be quite stark (see below).
- It is not well adapted to the sphere: the current method does not account for the curvature of the sphere in the interpolation and interpolates bilinearly along straight lines in the plane, as opposed to great circle geodesics.

### Describe the solution you'd like

I propose that the current method be supplemented by a more modern alternative developed for interpolation on the sphere, such as the [spherical mean value](https://www.mn.uio.no/math/english/people/aca/michaelf/papers/gbc.pdf) interpolation described by M. Floater. This algorithm is

- Universal across convex spherical polygons: The same formulae can be applied to both the quad and triangle (and even pentagons, if that is ever an issue which needs addressing), without having to divide the polygon in such a way that C1 discontinuities might be introduced.
- Is invariant under the cycling of polygon vertices: The same result is obtained regardless of the order in which the mesh is read.
- Is well adapted to the sphere: In a preliminary test suite interpolating real spherical harmonics up to 4th order on the cubed sphere, spherical mean value coordinates were found to be 33% more accurate than the current interpolation method, since they account for the curvature of the underlying surface.

Note that, while spherical mean value coordinates are more accurate at interpolating varying functions on the sphere, they are not suitable for interpolating uniform fields, where values at each of the cell vertices are equal. This is due to the weights not satisfying the partition of unity constraint, causing the curvature to dominate and the interpolant to increase near the centre of the cell. As such, the weights would need to be normalised or the existing method used in these cases.

### Describe alternatives you've considered

Other interpolation methods, such as (spherical/planar) Wachspress coordinates have been considered, however if the goal is increased flexibility and accuracy, spherical mean value coordinates seem to be the most suitable.

[...]

PR #329, opened 2025-11-05
Polygon implementation for modern (spherical mean value) interpolation.

lewisn-met:
### Description

This PR relates to issue ecmwf/atlas#328 and adds:

- A SphericalPolygon3D class as an alternative interpolation method to the Triag3D and Quad3D classes currently used in the finite-element interpolation method (`FiniteElement.h`). This computes weights using the spherical mean value formula rather than the bilinear/barycentric formulae currently implemented.

- A test suite for the new class, in `test_SphericalPolygon3D.cc`, which validates that the interpolation method correctly computes the areas, normals, and weights of an example triangle, planar quadrilateral, and non-planar quadrilateral on the surface of the unit sphere.

(Draft) PR integrating this class into the existing set of interpolation methods in atlas to follow, with the exact implementation to be fixed after further discussion with myself and @odlomax.  

### Contributor Declaration

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md).

### DIFF
--- a/src/atlas/util/ConvexSphericalPolygon.h
+++ b/src/atlas/util/ConvexSphericalPolygon.h
@@ -86,7 +86,7 @@ public:
 
     size_t size() const { return size_; }
 
-    std::vector<PointXYZ>& edge_normals() const {
+    const std::vector<PointXYZ>& edge_normals() const {
         if (not computed_edge_normals_) {
             compute_edge_normals();
         }
--- a/src/tests/util/test_convexsphericalmeanvalue.cc
+++ b/src/tests/util/test_convexsphericalmeanvalue.cc
@@ -80,11 +80,11 @@ CASE("test_convex_spherical_polygon_triag") {
         size_t pointsInside  = 0;
         size_t pointsOutside = 0;
 
-        for (size_t i = 0; i < numberTestPoints; ++i) {
-            std::vector<double> polygonWeights(testTriangle.size());
+        std::vector<double> polygonWeights(testTriangle.size());
 
+        for (size_t i = 0; i < numberTestPoints; ++i) {
             if (testTriangle.compute_vertex_weights(candidatePoints[i], polygonWeights.data(), polygonWeights.size()) ==
-                1) {
+                0) {
                 EXPECT((isPointInside[i] == 0));
                 pointsOutside += 1;
             }
@@ -161,11 +161,11 @@ CASE("test_spherical_polygon_nonplanar_quad") {
         size_t pointsInside  = 0;
         size_t pointsOutside = 0;
 
-        for (size_t i = 0; i < numberTestPoints; ++i) {
-            std::vector<double> polygonWeights(testQuad.size());
+        std::vector<double> polygonWeights(testQuad.size());
 
+        for (size_t i = 0; i < numberTestPoints; ++i) {
             if (testQuad.compute_vertex_weights(candidatePoints[i], polygonWeights.data(), polygonWeights.size()) ==
-                1) {
+                0) {
                 EXPECT((isPointInside[i] == 0));
                 pointsOutside += 1;
             }
@@ -178,7 +178,7 @@ CASE("test_spherical_polygon_nonplanar_quad") {
                 }
             }
         }
-        
+
         Log::info() << "Points in/out: " << pointsInside << "/" << pointsOutside << std::endl;
         EXPECT(pointsOutside == expectedOutside);
         EXPECT(pointsInside == expectedInside);
@@ -192,11 +192,11 @@ CASE("test_spherical_polygon_nonplanar_quad") {
         size_t pointsInsideRotated = 0;
         size_t pointsOutsideRotated = 0;
 
-        for (size_t i = 0; i < numberTestPoints; ++i) {
-            std::vector<double> polygonWeightsRotated(testQuadRotated.size());
+        std::vector<double> polygonWeightsRotated(testQuadRotated.size());
 
+        for (size_t i = 0; i < numberTestPoints; ++i) {
             if (testQuadRotated.compute_vertex_weights(candidatePoints[i], polygonWeightsRotated.data(),
-                                                       polygonWeightsRotated.size()) == 1) {
+                                                       polygonWeightsRotated.size()) == 0) {
                 EXPECT((isPointInside[i] == 0));
                 pointsOutsideRotated += 1;
             }


lewisn-met:
On a slightly unrelated note, I notice that the linux intel-classic ci and gnu-14 ci are are currently failing for this PR. These failures seem to be related to the pluto-benchmark and either the SphericalVector or Cubic2D classes. Is this something you'd like me to take a look at?

wdeconinck:
I have followed up on my hunch, and reverting the CI to use "eigen@3" seems to fix the compilation.
I have added this fix to develop and a 0.44.1 hotfix tag.

This is not a proper fix for compatibility Eigen 5.0.0 but perhaps these are infancy issues that will be sorted upstream in Eigen in due course. At least we can work without failing CI now.
Please rebase on develop.

lewisn-met:
I have now merged the fix from develop into this PR. Thanks for following your hunch so promptly - fingers crossed that any issues with Eigen 5. get sorted out soon!

I also noticed that the validate method included in your ConvexSphericalPolygon class nicely verifies the polygon is on the sphere and points are not too close. I could not see anywhere in the [FiniteElement](https://github.com/ecmwf/atlas/blob/358bfaf6988aa1e90586546cab2c1fcdf70f1ca4/src/atlas/interpolation/method/unstructured/FiniteElement.h#L29) class that does a similar check, so it may be worth sharing this functionality between the methods.

sbrdar:
Hi @lewisn-met. Many thanks for this interesting development. It seems to me that `SphericalPolygon3D` can be class-derived from `ConvexSphericalPolygon`. Can we assume that `SphericalPolygon3D` has to be convex spherical polygon?

lewisn-met:
Hi @sbrdar, thank you for taking a look at this. Indeed, we can assume `SphericalPolygon3D` is convex, though it need not be (the weights are still positive for star-shaped polygons and well-defined but negative otherwise). 

With regards to merging functionality, incorporating the methods of `SphericalPolygon3D` into `ConvexSphericalPolygon,` either directly or via inheritance, is certainly an option. The main differences/obstacles here are the representation of the polygon (`SphericalPolygon3D` requires the triple products from your `inLeftHemisphere()` function to be stored in addition to the vertices) and encapsulating what is actually necessary for each use case.

Hopefully we can discuss this soon.

sbrdar:
Hi @lewisn-met I am thinking that the Floater's barycentre could be used for the 2nd order cons. remapping in Atlas, so I cherry-picked your implementation with the unit test on SphericalVector3D, and added it to ConvexSphericalPolygon, see the usage [here](https://github.com/ecmwf/atlas/blob/0c17e5c6c6c4730478512a68ef7952473aa41b6c/src/tests/interpolation/test_SphericalPolygon3D.cc#L61). Maybe something to discuss soon.

lewisn-met:
Would incorporating this method into the cons. remapping be in addition or as an alternative to using Floater's interpolation in the finite-element class? 

Our original motivation for the algorithm was as a drop-in alternative to the Quad3D and Triag3D classes used for "continuous" interpolation, see the implementation [here](https://github.com/JCSDA-internal/atlas/blob/feature/spherical-mean-value-interpolation/src/atlas/interpolation/method/unstructured/FiniteElement.cc#L643). However, I can see how Floater's ideas could be useful in both settings and I don't expect the use of ConvexSphericalPolygon in FiniteElement.cc to be too different from the branch I have linked.

src/tests/interpolation/test_interpolation_cubedsphere.cc
```diff
@@ -84,26 +92,42 @@ CASE("cubedsphere_to_cubedsphere_interpolation") {
...
-    const auto scheme = util::Config("type", "cubedsphere-bilinear");
-    const auto interp = Interpolation(scheme, fixture.sourceFunctionSpace_, targetFunctionSpace);
+    for (util::Config config : interpolationTypes) {
+        std::string interpType        = "";
+        std::string normalisationMode = "";
+
+        config.get("type", interpType);
+        config.get("normalisation", normalisationMode);
+
+        SECTION("using " + interpType + " " + normalisationMode) {
```
sbrdar: A tiny comment here. This output can return, for instance, "using spherical-mean-value true" or "using spherical-mean-value false", but 'false'/'true' refers to normalisation. Can we have `normalisationMode` replaced with `(normalisationMode ? "w/ normalisation" : "")`?

src/tests/interpolation/test_interpolation_cubedsphere.cc
```diff
@@ -131,56 +155,70 @@ CASE("cubedsphere_scalar_interpolation") {
...
-    interp.execute(sourceField, targetField);
-    targetField.haloExchange();
+    for (util::Config config : interpolationTypes) {
+        std::string interpType        = "";
+        std::string normalisationMode = "";
+
+        config.get("type", interpType);
+        config.get("normalisation", normalisationMode);
+
+        SECTION("using " + interpType + " " + normalisationMode) {
```
sbrdar: A similar comment as before also here. This output can return, for instance, "using spherical-mean-value true" or "using spherical-mean-value false", but 'false'/'true' refers to normalisation. Can we have `normalisationMode` replaced with `(normalisationMode ? "w/ normalisation" : "")`?

src/tests/interpolation/test_interpolation_cubedsphere.cc
```diff
@@ -131,56 +155,70 @@ CASE("cubedsphere_scalar_interpolation") {
...
-            partView(i)    = mpi::rank();
-        }
-    }
-    partField.haloExchange();
+            if (normalisationMode != "") {
+                normalisationMode = "_" + normalisationMode;
+            }
 
-    gmshOutput("cubedsphere_source.msh", FieldSet(sourceField));
+            gmshOutput("cubedsphere_source_" + interpType + normalisationMode + ".msh", FieldSet(sourceField));
```
sbrdar: A tiny suggestion. Perhaps one can replace `normalisationMode` with `(normalisationMode ? "_normalised" : "")`

src/tests/interpolation/test_interpolation_cubedsphere.cc
```diff
@@ -131,56 +155,70 @@ CASE("cubedsphere_scalar_interpolation") {
...
-    auto targetFields = FieldSet{};
-    targetFields.add(targetField);
-    targetFields.add(errorField);
-    targetFields.add(partField);
-    gmshOutput("cubedsphere_target.msh", targetFields);
+            auto targetFields = FieldSet{};
+            targetFields.add(targetField);
+            targetFields.add(errorField);
+            targetFields.add(partField);
+            gmshOutput("cubedsphere_target_" + interpType + "_" + normalisationMode + ".msh", targetFields);
```
sbrdar: Here as well. A tiny suggestion. Perhaps one can replace `normalisationMode` with `(normalisationMode ? "_normalised" : "")`

src/tests/interpolation/test_interpolation_finite_element.cc
```diff
@@ -43,37 +51,47 @@ CASE("test_interpolation_finite_element") {
...
-        std::string test_string = test_stream.str();
-        EXPECT((test_string.find("max_fraction_elems_to_try: 0.4") != std::string::npos));
-    }
+        config.get("type", interpType);
+        config.get("normalisation", normalisationMode);
 
-    SECTION("test interpolation outputs") {
-        Field field_source = fs.createField<double>(option::name("source"));
-        Field field_target("target", array::make_datatype<double>(), array::make_shape(pointcloud.size()));
+        SECTION("using " + interpType + " " + normalisationMode) {
```
sbrdar: A tiny suggestion. Perhaps one can replace `normalisationMode` with `(normalisationMode ? "_normalised" : "")`

lewisn-met:
@sbrdar I see there is an issue with the preview-publish/deploy action - can you confirm that this is an issue on your end and not with this PR itself? It appears to be related to the python authenticator being called by the job runner.

On a separate note, would you like me to merge from develop, since I see there are several commits which have not yet made their way onto this branch?

sbrdar:
@lewisn-met Thanks very much for the merger with ConvexSphericalPolygon - great work! The tests pass correctly. The failing unrelated static_analyzer problem we can ignore for now and continue with the merge. Can you in the very last step merge from the latest develop?

lewisn-met:
@sbrdar many thanks for your kind words, I'm glad you like the work! The latest ECMWF/develop has now been merged into this branch and we are good to go.

I have a few small improvements for the spherical mean-value interpolator in mind, but will raise those in a separate PR when I get the chance to work on it. Thanks again for all of your help!

src/atlas/util/ConvexSphericalPolygon.cc
```diff
@@ -150,6 +169,65 @@ void ConvexSphericalPolygon::compute_centroid_and_area() const {
     computed_centroid_ = true;
 }
 
+// cf. M. Floater, “Generalized barycentric coordinates and applications” Acta Numerica, p. 001, 2016.
+std::optional<std::vector<double>> ConvexSphericalPolygon::compute_vertex_weights(
```
wdeconinck: Why optional?
Since it is a dynamically sized vector, would it not be same when just passing an empty vector in the "not present" case?
And then in the client code check for `vertex_weights.empty()`

src/atlas/util/ConvexSphericalPolygon.cc
```diff
@@ -117,8 +128,16 @@ ConvexSphericalPolygon::ConvexSphericalPolygon(const PointXYZ points[], size_t s
     else {
         ++isp;
     }
-    size_  = isp;
-    validate();
+    size_ = isp;
+    ATLAS_ASSERT(size_ > 2, "Polygon must have at least 3 points");
+    switch (mode) {
+        case SMV:
+            validateAndComputeNormals();
```
wdeconinck: `validateAndComputeNormals()` cannot  be split into the following, reusing the already available validate?
```c++
validate();
compute_edge_normals();
```

src/atlas/util/ConvexSphericalPolygon.h
```diff
@@ -67,14 +74,18 @@ class ConvexSphericalPolygon {
     using contains_PointLonLat = std::is_same<typename std::decay<typename Points::value_type>::type, PointLonLat>;
 
     template <class Points, typename std::enable_if<contains_PointLonLat<Points>::value, void>::type* = nullptr>
-    ConvexSphericalPolygon(const Points& points): ConvexSphericalPolygon(points.data(), points.size()) {}
+    ConvexSphericalPolygon(const Points& points, const interpolationMode mode = interpolationMode::GRID):
```
wdeconinck: The interpolationMode to me seems fishy.
What is fundamentally different depending on the mode?
The computed area or centroid do not seem different, so why the mode? I would really like to avoid this.

src/atlas/util/ConvexSphericalPolygon.h
```diff
@@ -86,6 +97,8 @@ class ConvexSphericalPolygon {
 
     size_t size() const { return size_; }
 
+    std::vector<PointXYZ> edge_normals() const { return edge_normals_; }
```
wdeconinck: This can be computed lazily on demand, just like the area.
Perhaps that could avoid the use of mode.

src/atlas/util/ConvexSphericalPolygon.h
```diff
@@ -130,6 +143,10 @@ class ConvexSphericalPolygon {
 
     int next(const int index) const { return (index == size_ - 1) ? 0 : index + 1; };
 
+    int previous(const int index) const { return (index == 0) ? size_ - 1 : index - 1; };
+
+    std::optional<std::vector<double>> compute_vertex_weights(const PointXYZ& candidatePoint) const;
```
wdeconinck: Another API may be desired and more performant, avoiding dynamic allocations:
```c++
int compute_vertex_weights(const PointXYZ& point, double vertex_weights[], size_t vertex_weights_size) {
  ATLAS_ASSERT(vertex_weights_size == size());
  ...
  return return_code; // instead of optional
}
```
The intended use would then be compatible with many data container types where the control of allocation is then on the client side.
```c++
std::vector<double> vertex_weights;
vertex_weights.reserve(max_size); // allocate outside of a loop
for (auto polygon : polygons ) {
    vertex_weights.resize( polygon.size() ); // no reallocation happening
    if (polygon.compute_vertex_weights( point, vertex_weights.data(), vertex_weights.size() ) == OK){
        // success, use vertex_weights
    }
}
```
or  example with std::span (C++20):
```c++
std::array<double,1000> buffer;
for (auto polygon : polygons ) {
    std::span<double> vertex_weights(buffer.data(), polygon.size());
    if (polygon.compute_vertex_weights( point, vertex_weights.data(), vertex_weights.size() ) == OK) {
        // success, use vertex_weights
    }
}
```

wdeconinck (changes requested):
Hello @lewisn-met I think this is looks a good merge of your previous polygon and this one.
I have some questions and suggestions.

src/atlas/util/ConvexSphericalPolygon.cc
wdeconinck: Actually ignore this, and see below for other suggested API, which I added as a review comment later.

src/tests/interpolation/test_interpolation_cubedsphere.cc
lewisn-met: Ah, well spotted! I agree "using spherical-mean-value true" would likely be a bit misleading to the user and have made this change (as well as those below) in a new commit. 

Some trickery was needed to interpret `normalisationMode` as a bool (since config.get outputs to a string), but hopefully nothing problematic.

src/atlas/util/ConvexSphericalPolygon.cc
lewisn-met: It is certainly possible to have the computation of the normals done in this way and I would be happy to do so if you would prefer.

The motivation for not doing so was because the GreatCircleSegments used for the validation are the same ones used for computing the edge normals. At present, these are thrown away in the `validate()` method, while they are used for both calculations in the `validate_edge_normals()` method. Essentially, it's a question of duplicate code vs duplicate constructor calls.

src/atlas/util/ConvexSphericalPolygon.h
lewisn-met: The difference between modes is whether or not the edge_normals are computed the validation step - this way we only compute and store the edge normals when the polygons are constructed for SMV interpolation and not for grid-to-grid conservative interpolation (where we want to save memory). 

An alternative approach, if we separate `validate()` and `compute_edge_normals()`, would be to have a call to `compute_edge_normals()` either before or at the start of each call to `compute_vertex_weights()`. Looking at your comment about lazy-computation below, I think this is the best option.

src/atlas/util/ConvexSphericalPolygon.h
lewisn-met: See above

src/atlas/util/ConvexSphericalPolygon.h
lewisn-met: I shall have a look at this idea and get back to you once I have done some testing. I certainly like the idea of avoiding dynamic allocation and adding compatibility features.

src/atlas/interpolation/method/unstructured/SphericalMeanValue.cc
```diff
@@ -0,0 +1,426 @@
...
+                (*icoords_)(idx[i], size_t(0)), (*icoords_)(idx[i], size_t(1)), (*icoords_)(idx[i], size_t(2))}));
+        }
+
+        util::ConvexSphericalPolygon currentPolygon(listVertices.data(), listVertices.size());
+
+        nb_cols = currentPolygon.size();
+        ATLAS_ASSERT(nb_cols == 3 || nb_cols == 4);
+
+        std::vector<double> polygonWeights;
+        polygonWeights.reserve(nb_cols);
```
wdeconinck:
```suggestion
        std::vector<double> polygonWeights(nb_cols);
```
`reserve` allocates but `polygonWeights.size()` remains zero, so this should not work, strange it would...
We can just use constructor.

src/atlas/util/ConvexSphericalPolygon.h
```diff
@@ -86,6 +87,13 @@ class ConvexSphericalPolygon {
 
     size_t size() const { return size_; }
 
+    std::vector<PointXYZ> edge_normals() const {
```
wdeconinck:
```suggestion
    const std::vector<PointXYZ>& edge_normals() const {
```
Avoid copy

src/atlas/util/ConvexSphericalPolygon.h
```diff
@@ -10,6 +10,7 @@
 
 #pragma once
 
+#include <optional>
```
wdeconinck: No longer needed.

src/atlas/util/ConvexSphericalPolygon.cc
```diff
@@ -92,8 +93,10 @@ ConvexSphericalPolygon::ConvexSphericalPolygon(const PointLonLat points[], size_
     else {
         ++isp;
     }
-    size_  = isp;
+    size_ = isp;
+    ATLAS_ASSERT(size_ > 2, "Polygon must have at least 3 points");
```
wdeconinck: This Assertion should not be needed,
`validate()` just after sets `valid_ = size_ > 2;`
The assertion is then on the user.

src/atlas/util/ConvexSphericalPolygon.cc
```diff
@@ -117,8 +120,10 @@ ConvexSphericalPolygon::ConvexSphericalPolygon(const PointXYZ points[], size_t s
...
```
wdeconinck: Same: This Assertion should not be needed,
validate() just after sets valid_ = size_ > 2;
The assertion is then on the user.

src/atlas/util/ConvexSphericalPolygon.cc
lewisn-met: Ah you are quite right! This was a hold-over from when `validateAndComputeNormals()` was still included in the class. I have removed this now.

src/atlas/interpolation/method/unstructured/SphericalMeanValue.cc
lewisn-met: Weirdly, I actually received a notification that the CI failed, but now it seems fine... very bizarre. In any case, you are right, this shouldn't work and I simply missed the call to `resize().` As you say though, using the constructor is the better solution.

src/atlas/util/ConvexSphericalPolygon.h
```diff
@@ -87,7 +86,7 @@ class ConvexSphericalPolygon {
 
     size_t size() const { return size_; }
 
-    std::vector<PointXYZ> edge_normals() const {
+    std::vector<PointXYZ>& edge_normals() const {
```
wdeconinck: You forgot to prefix with `const`
`const std::vector<PointXYZ>& edge_normals() const {`

src/tests/util/test_convexsphericalmeanvalue.cc
```diff
@@ -0,0 +1,227 @@
...
+            std::vector<double>{0, 0, 0, 0},
+            std::vector<double>{0, 0, 0, 0},
+            std::vector<double>{0, 0, 0, 0}  //
+        };
+
+        size_t pointsInside  = 0;
+        size_t pointsOutside = 0;
+
+        for (size_t i = 0; i < numberTestPoints; ++i) {
+            std::vector<double> polygonWeights(testQuad.size());
```
wdeconinck:
```suggestion
        std::vector<double> polygonWeights(testQuad.size());
        for (size_t i = 0; i < numberTestPoints; ++i) {
```
Move the allocation out of loop; not that it matters for performance in this test, but serves as example.

lewisn-met:
Many thanks for your comments @sbrdar @wdeconinck! I'm currently in the process of making the last few changes and debugging the ctest which now fails (unfortunately we are having some local build issues so this is taking longer than expected). I'll let you know once this is done.

wdeconinck (approved):
Thank you @lewisn-met for addressing all suggestions! This looks great now and good to merge.

lewisn-met:
That sounds great! From my side, I'm happy to go ahead with the merge - presumably the static-analyzer issue is related to the backend used for the checks/CI and not the PR itself?

PR #352, opened 2026-03-02
Fixes to pluto and hic

wdeconinck:
With `ECBUILD_2_COMPAT=ON` we were seeing errors in ifs-bundle like:
 ```
CMake Error: install(EXPORT "atlas-targets" ...) includes target "atlas" which requires target "hic" that is not in any export set.
CMake Error: install(EXPORT "atlas-targets" ...) includes target "atlas" which requires target "hicsparse" that is not in any export set.
CMake Error: install(EXPORT "pluto-targets" ...) includes target "pluto" which requires target "hic" that is not in any export set.
 ```
The culprit seems to be that in ecbuild_add_library, with `ECBUILD_2_COMPAT=ON`, INTERFACE libraries are not added to the `${PROJECT_NAME}_ALL_LIBS` variable which is used in `ecbuild_install_project`.
This PR manually adds hic and hicsparse to the list hic_ALL_LIBS to get exported, as a workaround.

Further fix includes adding a missing return value to `pluto%has_registered_resource`.

https://sites.ecmwf.int/docs/atlas/static-analyzer/PR-352

### DIFF
--- a/hic/CMakeLists.txt
+++ b/hic/CMakeLists.txt
@@ -74,6 +74,13 @@ ecbuild_add_resources(
         LICENSE
 )
 
+if (ECBUILD_2_COMPAT)
+  # Manually add hic and hicsparse to the list hic_ALL_LIBS to get exported.
+  # Only when ECBUILD_2_COMPAT=ON, INTERFACE libraries like hic,hicsparse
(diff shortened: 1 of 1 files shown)


wdeconinck:
Merged in develop

PR #353, opened 2026-03-05
Add ability to redistribute from NodeColumns to StructuredColumns

lewisn-met:
### Description

This PR adds the ability to construct a heterogeneous redistribution, that is, a redistribution between function spaces which are not of the same type. This functionality is accessed by specifying `util::Config("heterogeneous_distribution", "true")` in the Redistribution constructor's config argument. 

Furthermore, we currently restrict (using `ATLAS_ASSERT_MSG`) to the case where the sourceFunctionspace is a NodeColumns and the targetFunctionspace is a StructuredColumns, since this is the case required when performing an interpolation from a Cubedsphere (NodeColumns) to a Gauss (StructuredColumns), where a Gauss (NodeColumns) is taken as an intermediate step. 

We also provide a ctest which confirms that this new functionality performs the correct redistribution of double (ranks 1, 2, 3), float, int, and long fields and that exceptions are thrown when attempting to perform a heterogeneous redistribution with the default config option (none) or with function spaces of types other than those specified above.

### Contributor Declaration

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md).


src/tests/redistribution/test_redistribution_heterogeneous.cc
```diff
@@ -0,0 +1,285 @@
...
+                EXPECT(checkValue(this->targetView_(i, j, 0),
+                                  testPattern<Value>(targetLonlatView(i, LON), targetLonlatView(i, LAT), j)));
+                ++nCheck;
+                EXPECT(checkValue(this->targetView_(i, j, 1),
+                                  -testPattern<Value>(targetLonlatView(i, LON), targetLonlatView(i, LAT), j)));
+                ++nCheck;
+            }
+        }
+        const auto& comm = mpi::comm(this->sourceFunctionSpace_.mpi_comm());
+        comm.allReduceInPlace(nCheck, eckit::mpi::Operation::SUM);
```
wdeconinck:
```suggestion
        comm.allReduceInPlace(nCheck, eckit::mpi::sum());
```

wdeconinck (changes requested):
Hello @lewisn-met 
I would not like to introduce the extra option, and just change the logic to fail for non-compatible function spaces. Before this PR the requirement was to be the same type, but we can just remove this line 23:
```
ATLAS_ASSERT(sourceFunctionSpace.type() == targetFunctionSpace.type());
```
It can be up to the concrete implementations to make further restrictions if needed, but perhaps we don't need to make any restrictions. I can imagine that perhaps you want to redistribute with PointCloud, or perhaps StructuredColumns with CellColumns (case of healpix mesh)
.

odlomax:
> Hello @lewisn-met I would not like to introduce the extra option, and just change the logic to fail for non-compatible function spaces. Before this PR the requirement was to be the same type, but we can just remove this line 23:
> 
> ```
> ATLAS_ASSERT(sourceFunctionSpace.type() == targetFunctionSpace.type());
> ```
> 
> It can be up to the concrete implementations to make further restrictions if needed, but perhaps we don't need to make any restrictions. I can imagine that perhaps you want to redistribute with PointCloud, or perhaps StructuredColumns with CellColumns (case of healpix mesh) .


I'm more than happy for the safety to be left off, so to speak. If it's simplest to just remove the check, let's do that.

lewisn-met:
> Hello @lewisn-met I would not like to introduce the extra option, and just change the logic to fail for non-compatible function spaces... 
It can be up to the concrete implementations to make further restrictions if needed, but perhaps we don't need to make any restrictions. 

Hi @wdeconinck, thanks for taking a look at this. I'm ok with the idea of simply removing the assertion that the source and target are of the same type and allowing the user to perform redistribution between any two function spaces. It looks like the generic redistribution implementation is used by default and the ctest shows a heterogeneous choice works in this case (while RedistributionStructuredColumns fails for obvious reasons).

I don't know whether you would like some further tests for other choices added (such as the StructuredColumns to CellColumns you mentioned)? If so, I can take a look at adding these when I get the chance. Since these no longer require an additional parameter, I can probably move these into the tests for the generic redistribution.

wdeconinck (approved):
Looks good, sorry for the wait!

lewisn-met:
No problem, thanks for the review and for getting this merged!

PR #381, opened 2026-06-10
[python] pmr on

tmi:
Workaround for the wheel build

https://sites.ecmwf.int/docs/atlas/static-analyzer/PR-381

### DIFF
--- a/python/atlaslib-ecmwf/buildconfig
+++ b/python/atlaslib-ecmwf/buildconfig
@@ -11,7 +11,7 @@
 # TODO we duplicate information -- pyproject.toml's `name` and `packages` are derivable from $NAME and must stay consistent
 
 NAME="atlas"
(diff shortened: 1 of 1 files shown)


wdeconinck:
A bit more context after offline discussions:
The issue is that ENABLE_PMR=OFF (the default at this time) uses an internal replacement for the c++ STL library's `<memory_resource>` :  'compat/memory_resource' header without `.h` extension. The build_wheel infrastructure does not handle files without extension.
Two options:
- add `.h` extension
- rely on ENABLE_PMR=ON

The reason for the PMR option is that some systems have have a faulty C++17 installation where the `<memory_resource>` header is not found.
This PR opted for ENABLE_PMR=ON which is fine as the systems where the build_wheels infrastructure happens is tightly controlled to be C++17 compliant.
Downstream users of `pip install ...` are not supposed to use the headers.

wdeconinck:
NOTE, this solution seemed insufficient for some situations, see #382 , which reverts this PR.

PR #382, opened 2026-06-10
Fix python build_wheels issue with PMR for incompatible platforms

wdeconinck:
Revert the solution in #381 where the build_wheels tooling did not install the `pluto/src/pluto/memory_resource/compat/memory_resource` file due to missing `.h` extension.
The solution there counted on the fact that the `std::pmr` `memory_resource` header could be found. This is not the case on all containers/platforms where the build_wheels tooling is called. In particular following:
 `-- The C compiler identification is AppleClang 14.0.3.14030022`

**Compatibility Layer Improvements:**

* Added a new `README.md` in `pluto/src/pluto/memory_resource/compat/` explaining the purpose and usage of the `std::pmr` compatibility layer, including namespace selection and CMake configuration details.
* Renamed the compatibility header from `memory_resource` to `memory_resource.h` so that the build_wheels tooling can digest.

**Build and Configuration Updates:**

* Revert #381 , so removing `-DENABLE_PMR=1` from the build_wheels configuration.
* Added a developer note in `pluto/CMakeLists.txt` reminding maintainers to update the compatibility layer documentation if the default for `PMR` changes.

https://sites.ecmwf.int/docs/atlas/static-analyzer/PR-382


tmi:
looks good from my PoV -- here's a successful build from this branch for all platforms https://github.com/ecmwf/python-develop-bundle/actions/runs/27274546949. Thx a lot for this!

tmi (approved):

PR #395, opened 2026-07-29
Add a weighted space-filling-curve-based partitioner

### LINKED ISSUE #394: Add a weighted partitioner
### Is your feature request related to a problem? Please describe.

We run interpolations using atlas for model data to observation locations. These runs scale out fairly well to around 8 to 16 ranks, however beyond this we don't get much better performance and the MPI overhead gets larger. Looking at the results of benchmarking, the issue is a very large MPI-imbalance when applying the interpolation - some ranks have lots of work and others have very little.

The reasons for this are quite obvious. For example see the observation  density of sea surface temperature observations globally:

Image

Image

We don't currently have a partitioner that can adapt to increase the rank-density over ocean rather than land, or increase the rank-density where there is more work to do.

### Describe the solution you'd like

I think utilising the existing Hilbert curve construction to create a space-filling-curve partitioner is the most flexible, general solution for this. It is very easy to split the 1-D hilbert ordering into equal parts, or to weight each point in array by a weight and split by that weight.

### Describe alternatives you've considered

My first instinct was to experiment with already available partitioners, however none of them seemed to improve the issue.

The second approach is simply to increase the number of openMP threads per rank, however the sparse matrix multiply of the apply step doesn't scale well with threads past about 4. This is likely because of the cache-coherence of any matrix multiply.

### Options considered for load-balancing a grid partition

The options below were weighed by AI against how simple each is to build, how reusable it is beyond the ocean model, whether it actually balances observation load, its effect on halo/communication cost, and whether it needs new external data.

| Option | Simple to build | Reusable elsewhere | Balances obs load | Halo / comms cost | Needs new data | Notes |
|---|---|---|---|---|---|---|
| **A. Hard land/sea rank split** (separate rank pools for land and sea) | Medium | Low | **No** | Poor — two independent partitions meet at a ragged seam | No | Only recovers ranks wasted on all-land regions; an all-ocean pool still overloads northern ranks. Must pre-choose the land/sea rank split. |
| **B. Treat land as near-zero weight** (single pass) | High | Medium | Partly | Good | No | Strictly better than A. Stops spending ranks on land, but a binary land/sea weight still ignores the north/south observation imbalance. |
| **C. Weighted equal-area bands** (weight the existing band partitioner by obs density) | Medium | High | **Yes** | Good — keeps the equal-area band geometry | Yes (an obs-density climatology) | Targets the real cost and absorbs option B for free (land = zero obs). Requires editing the internals of the existing band partitioner and managing a climatology input. |
| **D. Weighted space-filling (Hilbert) curve** ⭐ *chosen* | **High** | **High** | **Yes** | Good — curve locality keeps each rank compact | Optional (weights are opt-in; unweighted still works) | Sort points along a Hilbert curve, then cut the line into equal-**weight** segments. Grid-agnostic, small self-contained addition, reuses an existing curve implementation, and the unweighted case is identical to plain point-count balancing. |
| **E. Industrial graph partitioner** (ParMETIS / Zoltan with vertex weights) | Low | High | **Highest** | Best — compact by construction | Yes | The "textbook" answer, but adds a heavy external dependency and build/integration cost. Overkill for a first step. |
| **F. Runtime-cost feedback** (re-balance from measured per-rank timings) | Low | High | Yes, adaptive | Good | Yes (a feedback loop) | Adapts to the actual run rather than a climatology, but needs runtime cost measurement and re-partitioning machinery. |
| **G. Analytic latitude weight** (no input files) | Highest | Medium | Weak | Good | No | e.g. a simple northern-hemisphere or cos-latitude boost. Zero new data — useful only as a crude first smoke-test of the plumbing. |

**Why option D (weighted space-filling curve):** it is the best balance of *simple to implement* and *powerful enough to fix the real imbalance*. The hard part — a correct, tested space-filling curve — already exists and is simply reused, so the new code is small and self-contained. It is **grid-agnostic and reusable** by any model with spatially varying cost, not just the ocean. Weighting is **opt-in**: with no weights it reproduces the standard point-count partition exactly, so it is a safe, strict generalisation rather than a replacement. Heavier approaches (E, F) can be revisited later if needed, but D delivers most of the benefit at a fraction of the cost and complexity.

### Additional context

_No response_

### Organisation

Met Office

twsearle:
### Description

Add a Hilbert space-filling-curve partitioner to atlas ("hilbert"), with optional per-point weighting.

Atlas' geometric partitioners balance point count across latitude bands. For grids where cost is spatially non-uniform (e.g. ORCA ocean grids where observations cluster in the northern hemisphere and vanish over land), this leaves MPI ranks badly imbalanced.

This adds a "hilbert" partitioner: points are sorted along a Hilbert curve and the 1-D ordering is cut into contiguous segments. The curve's spatial locality keeps each segment compact (low halo volume). By default the cut is equal-count and bit-identical to the existing equal-count result; an opt-in weight function (analytical, or a precomputed per-point array) instead balances total weight, so it's a strict generalisation.

### Issue Addessed

Fixes #394

### Summary of Changes

 *  The existing Hilbert-curve code is factored out of ReorderHilbert into a reusable grid::HilbertCurve helper (no algorithm change)
 *  atlas-meshgen gains --partition-points (a polygon-free, MPI-free scatter plot of the partition, useful for any partitioner) and a --weighted switch demonstrating the weight API.
 * Unit tests cover the equal-count invariant, the weight-balance property, the unit-weight (neutral) default, and the array-weight path.

### Testing

I have added unit tests, along with testing with the adapted `atlas-meshgen` executable.

With the O48 grid and the hilbert partitioner for 8 processors:

partition_points_unweighted

With the example 8x weight on northern hemisphere points, where the point size denotes the weight of each point:

partition_points_weighted

#### Ocean-specific testing

I setup a prototype partitioning of the extended ORCA025_T grid, using a weight distribution from *in-situ* observations in a testing branch of atlas-orca.

The weights are read from a simple 1-degree regular grid file:

partition_weights_weights

This is used in the weighted Hilbert space-filling curve partitioner to produce:

image

### Known limitations

 * May produce spatially discontinuous partitions. 
   * Unlike the equal-area/band partitioners, a space-filling-curve partition (especially when weighted, or across the periodic seam/poles) is not guaranteed to be a single connected region per rank.
   * Partition distributions and halo build/exchange do not require contiguity and are unaffected, but anything that relies on building a single simply-connected partition polygon can fail. e.g the lon/lat-polygon matching-mesh partitioner and polygon-based point location used by some interpolation setups.
   * Whether discontinuous partitions are acceptable everywhere across Atlas has not been exhaustively verified.
 * Not yet exercised in production. 
   * The partitioner has unit-test coverage but is not yet used by any downstream mesh/model, so it has not been validated across the full range of Atlas grids and mesh generators.
 * No HPC performance evaluation yet.
   * The load-balancing benefit for the motivating case (observation-driven imbalance) has not been measured.
 * Weighting is opt-in and caller-supplied.
   * Weights are set programmatically (not via the string/config factory), so config-driven runs get the equal-count behaviour only. 
   * This is intentional (the balancing logic stays domain-agnostic; the caller owns where weights come from), but it means weighting can't currently be enabled purely from YAML/CLI.
 * Partitioning is computed independently on every rank.
   * Each rank sorts the full point set (O(N log N), no communication). This matches the other geometric partitioners and keeps results deterministic.

### Contributor Declaration

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/Contributor-License-Agreement.md).

### DIFF
--- a/src/apps/atlas-meshgen.cc
+++ b/src/apps/atlas-meshgen.cc
@@ -20,6 +20,7 @@
 #include "atlas/functionspace/EdgeColumns.h"
 #include "atlas/functionspace/NodeColumns.h"
 #include "atlas/grid.h"
+#include "atlas/grid/detail/partitioner/SpaceFillingCurvePartitioner.h"  // for the --weighted demonstration
 #include "atlas/library.h"
 #include "atlas/mesh/Mesh.h"
 #include "atlas/mesh/Nodes.h"
@@ -106,6 +107,33 @@ Partitioner make_partitioner(const Grid& grid, const AtlasTool::Args& args) {
     if (args.has("partitions")) {
         config.set("partitions", args.getInt("partitions"));
     }
+
+    // ---------------------------------------------------------------------
+    // Demonstration of the Hilbert partitioner's opt-in weighting. The weight
+    // interface is deliberately not part of the string/config factory (real
+    // weights come from data, e.g. observation density), so we build the concrete
+    // partitioner here, set an illustrative weight, and wrap it in a Partitioner
+    // handle (which takes ownership). This lets users explore how weighting
+    // reshapes the partition without wiring up a data source.
+    if (args.getBool("weighted", false)) {
+        using grid::detail::partitioner::SpaceFillingCurvePartitioner;
+        std::string type;
+        config.get("type", type);
+        if (type == "hilbert") {
+            const double heavy = args.getDouble("weight-factor", 8.0);
+            auto* concrete     = new SpaceFillingCurvePartitioner(config);
+            // Northern-hemisphere-heavy demo weight (proxy for observation
+            // density): northern points cost `heavy` x southern points, so the
+            // curve is cut into segments of equal total weight, giving northern
+            // partitions proportionally fewer points.
+            concrete->set_weight_function(
+                [heavy](gidx_t, const PointXY& p) { return p.y() > 0. ? heavy : 1.0; });
+            return Partitioner{concrete};
+        }
+        Log::warning() << "--weighted ignored: only supported for --partitioner=hilbert" << std::endl;
+    }
+    // ---------------------------------------------------------------------
+
     return Partitioner{config};
 }
 
@@ -172,6 +200,16 @@ Meshgen2Gmsh::Meshgen2Gmsh(int argc, char** argv): AtlasTool(argc, argv) {
 
     add_option(new Separator("Options for `--partitioner=checkerboard`"));
     add_option(new SimpleOption<bool>("regular", "regular checkerboard partitioner"));
+
+    add_option(new Separator("Options for `--partitioner=hilbert` (visualisation / weighting demo)"));
+    add_option(new SimpleOption<bool>("weighted",
+                                      "Apply a demo northern-hemisphere-heavy weight to the hilbert partitioner"));
+    add_option(new SimpleOption<double>("weight-factor",
+                                        "Weight of northern vs southern points for --weighted (default = 8)"));
+    add_option(new SimpleOption<bool>(
+        "partition-points",
+        "Write a single-process python scatter plot (partition_points.py) of the whole-grid partition assignment. "
+        "Needs no MPI; use --partitions to set the number of partitions."));
 
     add_option(new Separator("Advanced"));
     add_option(new SimpleOption<long>("halo", "Halo size"));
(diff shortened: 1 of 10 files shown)


wdeconinck:
Hello @twsearle this looks like a nice feature and thank you for the thorough analysis. I have no objections to go further with this. Nice reuse of the existing Hilbert algorithm. I'm happy it serves you well.

As stated in your analysis there is the real problem of disconnected points with this partitioner, but that is only problematic for certain use cases. Some that I can foresee now are
- algorithms that compute or use a single polygon (for instance in the MatchingPartitioner)
- The "StructuredMeshGenerator" relies on a particular shape like the equal-regions partitioner.
Having full support for disconnected partitions is something that has been on my wish list for some time though, but is not in my near term plans because (until now?) no real motivation.

The concept of weighting is also very interesting. I think it could apply to other partitioners as well, and it could be worthwhile to come to a generalisation for the "Partitioner" concept. I would try to come up with a better solution for this before going ahead with this PR, unless you want to split this PR in two:
1. Just the new SpacefillingCurve (hilbert) partitioner without any notion of weights
2. Add weight handling in generic way with modifications to the SpacefillingCurve partitioner.

For 2. we could for instance create a global registry of weight_functions by name, which we should also be able to customise and add to at runtime. We could then pass the particular weight function into the configuration of the partitioners that support it. 
Conceptually:
```c++
double heavy{8};
std::vector<double> weights; // fill in yourself
Partitioner::register_weights("toby_weight_function", [heavy](gidx_t, const PointXY& p) { return p.y() > 0. ? heavy : 1.0; });
Partitioner::register_weights("toby_weights",         weights); // shorthand for your move of weights into a function
Partitioner with_weight_function{"hilbert", util::Config("weights", "toby_weight_function")};
Partitioner with_weights        {"hilbert", util::Config("weights", "toby_weights")};
...
// After use we can or should unregister no longer needed entries.
Partitioner::unregister_weights("toby_weight_function");
Partitioner::unregister_weights("weight_function");
```
In that case we now no longer need to first make a concrete partitioner and set the function, and we can create other partitioners in plugins seamlessly with weighting.

We need to be sure of the weight function signature. You added both `gidx_t` and `PointXY`. Could you elaborate on your choice here? Could you envision other dynamic arguments that play a role?

twsearle:
Hi @wdeconinck thanks yes! This has been a fun little side project to explore.

A bit more about my motivation - I was looking at some of our runs and I thought that the MPI imbalance was down to the fact that a lot of the observations in the target field for interpolation were landing on a small fraction of the MPI ranks. I was asking myself, what if I could make some sort of better compromise in the partitioner between the cost of compute due to observations vs the cost of compute due to the mesh.

What I ended up doing was exploring in atlas-orca a sparse meshgenerator that would work with this partitioner, and then a constructed weights file based on the observations. I did some experiments to try and parameterise these weights.

In the end the specific usecase I was exploring it turned out that the imbalance wasn't the observations at all! In fact there is a design decision in OOPS that leads to a lot of wasted constant-time work as the interpolator seems to be called Nrank times per rank. I found that a small change (#397) makes all the difference and eliminates a bunch of copy-cost, making all this inefficiency essentially invisible!

So although I had a fun time doing it, for now I kind of eliminated my use-case!

> In that case we now no longer need to first make a concrete partitioner and set the function, and we can create other partitioners in plugins seamlessly with weighting.
> 
> We need to be sure of the weight function signature. You added both `gidx_t` and `PointXY`. Could you elaborate on your choice here? Could you envision other dynamic arguments that play a role?

Having said all of the above, it is worth pointing out that the reason I didn't stick with just a function is that I wanted the ability to do something more complicated with the weights. I wanted to allow us to:
 - optimise them for a particular workload
 - handle patterns that are too complicated for a simple function (such as a changing per-workload land-sea mask)
 - allow the possibility of some future feedback mechanism (like how FFTW runs your problem a bunch of times, and can then cache the best strategy for running your particular problem)
 
Having said all that, I am happy to let this change go, break it down into just the hilbert partitioner for now or whatever you think best! The reason I decided to add in the weights to this PR as well was really just I was concerned that you would look at the generated partitions and think very reasonably "why do we need this? The result is achievable with existing partitioners". This partitioner makes the weighting thing super easy to implement, which was the real goal of me adding the partitioner in the first place.

wdeconinck:
Thanks @twsearle for the clarifications.
I suppose we can just have the partitioner without weights at the moment then, if you're up to it. And if the weights become important, we can look into that part again.

PR #397, opened 2026-08-05
Interpolation apply: gather only referenced source rows in the application of interpolation with missing values

twsearle:
### Summary

Improve performance of interpolation application with fields with missing values. Add similar machinery to the rank-3 interpolation as well (we will need this for future vector interpolation).

### Description

When an interpolation has a non-linear update (e.g `non_linear: missing-if-all-missing`) applied to a rank-2 (multi-level) source field, `Method::interpolate_field_rank2` copied the entire source column into a per-level temporary on every level of every apply:

```cpp
for (lev) for (i in 0..src.shape(0)) src_slice_v(i) = src_v(i, lev);
```

This copy is `O(source_size × levels)` and independent of the number of target points so it dominated `apply` for observation operators where the target is far smaller than the source.

Both the non-linear missing-value redistribution and the subsequent sparse matmul only ever read the source at the columns referenced by the weight matrix `W`. This PR precomputes the unique set of referenced source rows once, then copies only those rows per level, turning the copy into `O(nnz × levels) = O(target × stencil × levels)`. The result is bit-for-bit identical (unreferenced temporary entries were never read).

A cheap `nnz < source_size` guard falls back to the original full-column copy for regridding-like interpolations (target ≈ source, most rows referenced), where building the referenced set would be pure overhead.

**Performance**

Measured on a JEDI OOPS HofX4D application, using eORCA12 (~15M points), 75 levels, 32 MPI × 4 OMP:

- `apply` went from being the dominant, node-scaling bottleneck to 47.78 ms total across 64 calls (0.75 ms/call).
- total runtime for the test case went from over 8 mins to approximately 5 mins
- The obs-independent node term is eliminated; residual apply cost now scales only with observation count.

Measured on a regridding application, from eORCA025 to eORCA12 with 75 levels, 11 fields 8 MPI × 8 OMP:

- no measurable performance cost from adding the extra check

**Scope**

Single-file change in `src/atlas/interpolation/method/Method.cc`

### Contributor Declaration

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/Contributor-License-Agreement.md).

### DIFF
--- a/src/atlas/interpolation/method/Method.cc
+++ b/src/atlas/interpolation/method/Method.cc
@@ -8,7 +8,9 @@
  * nor does it submit to any jurisdiction.
  */
 
+#include <algorithm>
 #include <memory>
+#include <vector>
 
 #include "atlas/interpolation/method/Method.h"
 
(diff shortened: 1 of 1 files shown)


wdeconinck (approved):
Thanks @twsearle I have nothing to add, good optimisation!

